Issue #16, opened 2019-01-13
Speed of sum(BroadcastArray)


mcabbott:
Here's something I was trying to do recently, to avoid materializing a huge array just to sum it: 
```
V = rand(500);
V3 = reshape(V, 1,1,:);
@time sum(V .* V' .* V3)                                 # 0.57 seconds, 953.675 MiB
@time sum(BroadcastArray(*, V, V', V3))                  # 0.36 seconds
@time sum(BroadcastArray(*, V, V', V3), dims=(1,2,3))[1] # 0.014 seconds
```
Might it be worth making `sum()` simply call the `dims=...` version here?

Before I saw this package I was messing around with broadcasting to do this by hand, and could get in the ballpark of 0.014s here, but no better. Doing that seems likely to be more fragile.

dlfivefifty:
Happy to accept a PR fixing this!

mcabbott:
Will do. 

But first, one more puzzle which just bit me: why is the first of these 5x quicker? 
```
V2 = reshape(V, 1,:)
@time sum(BroadcastArray(*, V, V', V3), dims=(2,3));    # 0.025 seconds
@time sum(BroadcastArray(*, V, V2, V3), dims=(2,3));    # 0.132 seconds
```
Same times here: 
```
W = similar(V);
@time sum!(W, BroadcastArray(*, V, V', V3));
@time sum!(W, BroadcastArray(*, V, V2, V3));
```

dlfivefifty:
I looked at the profile and the second one spends more time in `getindex`. I can't seem to think of a good reason why.

mcabbott:
Thanks for taking a look, it is indeed strange. The biggest difference in an isolated getindex I can see is... far from a factor of 5: 
```
B1 = BroadcastArray(*, V, V', V3).broadcasted
B2 = BroadcastArray(*, V, V2, V3).broadcasted
@btime $B1[3,3,3]    # 3.816 ns
@btime $B2[3,3,3]    # 4.205 ns
```

dlfivefifty:
Remember order of access of data makes a big difference, but I’m not sure how to test that

mcabbott:
Yes that could be what's happening. I was trying to dig into what `CartesianIndices` etc. give, but couldn't find any difference. Here's another attempt, but I don't know what this means...
```
V = collect(11:12)/1;

W = collect(21:23)/1;
V2 = reshape(collect(21:23)/1, 1,:);

[...]

PR #21, opened 2019-02-06
Lazy broadcasting macro

mcabbott:
This PR adds a macro `@lazy` to turn broadcasting expressions into BroadcastArrays, and `@lazydot` which applies `@.` first. (It also replaces #18 with an explicit sum method.)

The clever idea here to hack broadcasting is due to @dawbarton and @tkf, discussed here: 

https://discourse.julialang.org/t/boolean-short-circuit-fusion/19641/20   
https://github.com/JuliaLang/julia/issues/19198#issuecomment-457967851   
https://github.com/JuliaLang/julia/pull/30939   

Perhaps this package would prefer an `@lazy` which also converted `vcat -> Vcat` etc, haven't thought about that. The name `@lazy` also clashes with ~~MacroTools~~ Lazy.jl; I was going to suggest `@z` & `@ż` as short names but can't seem to type `ż` at the REPL (on a mac)... how ugly are `@□` & `@⊡`?

### DIFF
--- a/src/lazybroadcasting.jl
+++ b/src/lazybroadcasting.jl
@@ -61,7 +61,7 @@ function Base._prod(f, A::BroadcastArray, ::Colon)
 	out
 end
 
-# Macros for lazy broadcasting
+# Macros for lazy broadcasting, #21 WIP
 # based on @dawbarton  https://discourse.julialang.org/t/19641/20
 # and @tkf            https://github.com/JuliaLang/julia/issues/19198#issuecomment-457967851
 # and @chethega      https://github.com/JuliaLang/julia/pull/30939
@@ -73,7 +73,7 @@ struct LazyCast{T}
     value::T
 end
 Broadcast.broadcasted(::typeof(lazy), x) = LazyCast(x)
-Base.materialize(x::LazyCast) = BroadcastArray(x.value)
+Broadcast.materialize(x::LazyCast) = BroadcastArray(x.value)
 
 lazyhelp = """
 	@lazy A .+ B     == @□ A .+ B
--- a/src/lazybroadcasting.jl
+++ b/src/lazybroadcasting.jl
@@ -66,58 +66,43 @@ end
 # and @tkf            https://github.com/JuliaLang/julia/issues/19198#issuecomment-457967851
 # and @chethega      https://github.com/JuliaLang/julia/pull/30939
 
-export @lazy, @lazydot, @□, @⊡
+export @~
 
-lazy(::Any) = error("function `lazy` must be called with a dot")
+lazy(::Any) = throw(ArgumentError("function `lazy` exists only for its effect on broadcasting, see the macro @~"))
 struct LazyCast{T}
     value::T
 end
 Broadcast.broadcasted(::typeof(lazy), x) = LazyCast(x)
 Broadcast.materialize(x::LazyCast) = BroadcastArray(x.value)
 
-lazyhelp = """
-    @lazy A .+ B     == @□ A .+ B
-    @lazydot A + B   == @⊡ A + B
-
-Macros for creating lazy `BroadcastArray`s: `@lazy` expects a broadcasting expression, 
-while `@lazydot` applies `@.` first. Short forms are typed `@\\square` and `@\\boxdot` 
-(or perhaps `@z` & `@ż` except that `ż` seems hard to enter at the REPL).
 """
+    @~ expr
 
-@doc lazyhelp
-macro lazy(ex)
-    checkex(ex)
-    :( lazy.($(esc(ex))) )
-end
-@doc lazyhelp
-macro □(ex)
-    checkex(ex)
-    :( lazy.($(esc(ex))) )
-end
+Macro for creating lazy `BroadcastArray`s. 
+Expects a broadcasting expression, possibly created by the `@.` macro:
+```
+julia> @~ A .+ B ./ 2
 
-@doc lazyhelp
-macro lazydot(ex)
-    checkex(ex, "@lazydot")
-    :( @. lazy($(esc(ex))) )
-end
-@doc lazyhelp
-macro ⊡(ex)
-    checkex(ex, "@lazydot")
-    :( @. lazy($(esc(ex))) )
+julia> @~ @. A + B / 2
+```
+"""
+macro ~(ex)
+    checkex(ex)
+    esc( :( $lazy.($ex) ) )
 end
 
 using MacroTools 
 
-function checkex(ex, name="@lazy")
+function checkex(ex)
     if @capture(ex, (arg__,) = val_ ) 
         if arg[2]==:dims
-            throw(ArgumentError("$name is capturing keyword arguments, try with `; dims = $val` instead of a comma"))
+            throw(ArgumentError("@~ is capturing keyword arguments, try with `; dims = $val` instead of a comma"))
         else
-            throw(ArgumentError("$name is probably capturing capturing keyword arguments, needs a single expression"))
+            throw(ArgumentError("@~ is probably capturing capturing keyword arguments, try with ; or brackets"))
         end
     end
     if @capture(ex, (arg_,rest__) ) 
-        throw(ArgumentError("$name is capturing more than one expression, try $name($arg) with brackets"))
+        throw(ArgumentError("@~ is capturing more than one expression, try $name($arg) with brackets"))
     end
     ex
 end
--- a/src/lazybroadcasting.jl
+++ b/src/lazybroadcasting.jl
@@ -37,9 +37,91 @@ getindex(B::BroadcastArray{<:Any,1}, kr::AbstractVector{<:Integer}) =
 
 copy(bc::Broadcasted{<:LazyArrayStyle}) = BroadcastArray(bc)
 
-# issue 16: sum(b, dims=(1,2,3)) faster than sum(b)
-Base._sum(b::BroadcastArray{T,N}, ::Colon) where {T,N} = first(Base._sum(b, ntuple(identity, N)))
-Base._prod(b::BroadcastArray{T,N}, ::Colon) where {T,N} = first(Base._prod(b, ntuple(identity, N)))
+# Replacement for #18.
+# Could extend this to other similar reductions in Base... or apply at lower level? 
+# for (fname, op) in [(:sum, :add_sum), (:prod, :mul_prod),
+#                     (:maximum, :max), (:minimum, :min),
+#                     (:all, :&),       (:any, :|)]
+function Base._sum(f, A::BroadcastArray, ::Colon)
+	bc = A.broadcasted
+	T = Broadcast.combine_eltypes(f ∘ bc.f, bc.args) 
+	out = zero(T)
+	@simd for I in eachindex(bc)
+	    @inbounds out += f(bc[I])
+	end
+	out
+end
+function Base._prod(f, A::BroadcastArray, ::Colon)
+	bc = A.broadcasted
+	T = Broadcast.combine_eltypes(f ∘ bc.f, bc.args) 
+	out = one(T)
+	@simd for I in eachindex(bc)
+	    @inbounds out *= f(bc[I])
+	end
+	out
+end
+
+# Macros for lazy broadcasting
+# based on @dawbarton  https://discourse.julialang.org/t/19641/20
+# and @tkf            https://github.com/JuliaLang/julia/issues/19198#issuecomment-457967851
+# and @chethega      https://github.com/JuliaLang/julia/pull/30939
+
+export @lazy, @lazydot, @□, @⊡
+
+lazy(::Any) = error("function `lazy` must be called with a dot")
+struct LazyCast{T}
+    value::T
+end
+Broadcast.broadcasted(::typeof(lazy), x) = LazyCast(x)
+Base.materialize(x::LazyCast) = BroadcastArray(x.value)
+
+lazyhelp = """
+	@lazy A .+ B     == @□ A .+ B
+	@lazydot A + B   == @⊡ A + B
+
+Macros for creating lazy `BroadcastArray`s: `@lazy` expects a broadcasting expression, 
+while `@lazydot` applies `@.` first. Short forms are typed `@\\square` and `@\\boxdot` 
+(or perhaps `@z` & `@ż` except that `ż` seems hard to enter at the REPL).
+"""
+
+@doc lazyhelp
+macro lazy(ex)
+	checkex(ex)
+    :( lazy.($(esc(ex))) )
+end
+@doc lazyhelp
+macro □(ex)
+	checkex(ex)
+    :( lazy.($(esc(ex))) )
+end
+
+@doc lazyhelp
+macro lazydot(ex)
+	checkex(ex, "@lazydot")
+    :( @. lazy($(esc(ex))) )
+end
+@doc lazyhelp
+macro ⊡(ex)
+	checkex(ex, "@lazydot")
+    :( @. lazy($(esc(ex))) )
+end
+
+using MacroTools 
+
+function checkex(ex, name="@lazy")
+	if @capture(ex, (arg__,) = val_ ) 
+		if arg[2]==:dims
+			throw(ArgumentError("$name is capturing keyword arguments, try with `; dims = $val` instead of a comma"))
+		else
+			throw(ArgumentError("$name is probably capturing capturing keyword arguments, needs a single expression"))
+		end
+	end
+	if @capture(ex, (arg_,rest__) ) 
+		throw(ArgumentError("$name is capturing more than one expression, try $name($arg) with brackets"))
+	end
+	ex
+end
+
 
 BroadcastStyle(::Type{<:BroadcastArray{<:Any,N}}) where N = LazyArrayStyle{N}()
 BroadcastStyle(L::LazyArrayStyle{N}, ::StaticArrayStyle{N}) where N = L
--- a/src/lazybroadcasting.jl
+++ b/src/lazybroadcasting.jl
@@ -43,22 +43,22 @@ copy(bc::Broadcasted{<:LazyArrayStyle}) = BroadcastArray(bc)
 #                     (:maximum, :max), (:minimum, :min),
 #                     (:all, :&),       (:any, :|)]
 function Base._sum(f, A::BroadcastArray, ::Colon)
-	bc = A.broadcasted
-	T = Broadcast.combine_eltypes(f ∘ bc.f, bc.args) 
-	out = zero(T)
-	@simd for I in eachindex(bc)
-	    @inbounds out += f(bc[I])
-	end
-	out
+    bc = A.broadcasted
+    T = Broadcast.combine_eltypes(f ∘ bc.f, bc.args) 
+    out = zero(T)
+    @simd for I in eachindex(bc)
+        @inbounds out += f(bc[I])
+    end
+    out
 end
 function Base._prod(f, A::BroadcastArray, ::Colon)
-	bc = A.broadcasted
-	T = Broadcast.combine_eltypes(f ∘ bc.f, bc.args) 
-	out = one(T)
-	@simd for I in eachindex(bc)
-	    @inbounds out *= f(bc[I])
-	end
-	out
+    bc = A.broadcasted
+    T = Broadcast.combine_eltypes(f ∘ bc.f, bc.args) 
+    out = one(T)
+    @simd for I in eachindex(bc)
+        @inbounds out *= f(bc[I])
+    end
+    out
 end
 
 # Macros for lazy broadcasting, #21 WIP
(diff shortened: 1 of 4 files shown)


dlfivefifty:
Great! Though I think we should think through the names of the macros carefully. Once `Applied` is merged, we have a few different "lazy" objects and so `@lazy` on its own may be confusing:

1. `broadcasted` in base. I don't think we need to worry about this one.
2. `BroadcastArray`
3. `applied` in LazyArrays.jl. Plays the same role as `broadcasted` but for lazy applying. E.g., `applied(exp, A)` is an un-realised matrix exponential.
4. `ApplyArray`. For example, `ApplyArray(exp, A)` represents the matrix exponential as an `AbstractMatrix`.

mcabbott:
Yes, this is just a first attempt, I was playing with macros and found myself in danger of re-inventing the wheel... better to use BroadcastArray. 

Agree the name is confusing, or too limited, in the context of the rest of this package. One could conceivably make `@lazy exp(A) == ApplyArray(exp, A)` and `@lazydot exp(A) == applied(exp, A)` but I'm not sure I volunteer to write such a thing. The more limited actually existing one could be called `@lazycast`? `@bc`? I don't know.

tkf:
How about using a supre short macro like `@~`?  It would be less distracting in an equation and we can "avoid" the naming problem.  Tilde kind of look like a wave and so I think it can mean "floating", i.e., non-materializing behavior.

> Once `Applied` is merged, we have a few different "lazy" objects

Any reason why for exposing both the expression trees (`broadcasted` and `applied`) _and_ array wrappers (`BroadcastArray` and `ApplyArray`) to the _end-users_ (i.e., not library writers extending LazyArray)?  If we don't immediately materialize the expression (e.g., `Y .= Mul(A, X)`) I think it's nicer to have an array-like object so that it can work nicely with the rest of Julia ecosystem.  As for `BroadcastArray` vs `ApplyArray`, I don't think we should choose one.  It's better if I can write `@lazy (A .* B) ⋆ C`.

src/lazybroadcasting.jl
```diff
@@ -37,9 +37,91 @@ getindex(B::BroadcastArray{<:Any,1}, kr::AbstractVector{<:Integer}) =
...
+
+Macros for creating lazy `BroadcastArray`s: `@lazy` expects a broadcasting expression, 
+while `@lazydot` applies `@.` first. Short forms are typed `@\\square` and `@\\boxdot` 
+(or perhaps `@z` & `@ż` except that `ż` seems hard to enter at the REPL).
+"""
+
+@doc lazyhelp
+macro lazy(ex)
+    checkex(ex)
+    :( lazy.($(esc(ex))) )
```
tkf: Isn't `esc(:( $lazy.($ex) ))` better?  It'd work better with macros inside `@lazy`.

tkf:
Another point: why do you need `@lazydot`?  Why not `@lazy @.`?

mcabbott:
Yes it could be shorter, I wondered about having a pair of symbols with & without a dot, but don't much like the ones I could think of above. I like the `@~` idea, I never write `~` so this suggests nothing else to me. Indeed there is no great need for a separate `@lazydot`.

dlfivefifty:
I like it! One could imagine ` ~ ` playing the equivalent role as ` . `, eg ` y ~= exp~(A) ` could in the future be in-place matrix exponential.

src/lazybroadcasting.jl
mcabbott: I never thought to use `$` for that, am new to macros. Both give `@macroexpand @~ @. A + 2` --> `:((LazyArrays.lazy).((+).(A, 2)))`, how do I see the difference?

tkf:
I should have mentioned that the downside of `~` is that it already means bitwise not.  If you are somehow using `~x` all the time, it may be confusing.  But it seems you guys like it so maybe I don't need to worry :)

mcabbott:
Well spotted, `Base.:~(f::Function, x::Any) = Applied(f,x)` seems to work.

dlfivefifty:
Let’s avoid inconsistency if we can. I think a Unicode symbol is better than overloading bitwise not.

src/lazybroadcasting.jl
tkf: Ah, sorry, never mind.  I think I was mixing it with something else.

tkf:
An ascii symbol that is available and not an operator is `$`.  So, we can use `@$` in principle.  There is also `@_`.  It would be nice if we can use `@:` but that's a parse error...

mcabbott:
I've never written `~` intentionally; someone who does would have to tell me if it ever happens around broadcasting. Other short names: `@_` (also downstairs?) `@¬` (also a negation, \neg) `@|` (barrier to materialisation) `@□` (open \square)? 

Anyway no rush, this is mostly to see if it functions well first. My tests like #16 are about large arrays, but perhaps other things (such as allocations of small broadcasts) may need more scrutiny.

tkf:
`@..` is also usable :)

dlfivefifty:
Ideally it would look "nice" as dot syntax. Not that it's going to be incorporated into the parser anytime soon, but just better to have it evoke it. `..` is OK:

```julia
y ..= exp..(A)
```
but I think it's more appropriate for a "tiered" broadcast (`broadcast(v -> broadcast(exp, v), A)`).

□ is OK:
```julia
y □= exp□(A)
```
but not great.

¬ is also OK:
```julia
y ¬= exp¬(A)
```
but knowing AppleScript I just see "new line".

_, |, and : are not good.

tkf:
> but I think it's more appropriate for a "tiered" broadcast (`broadcast(v -> broadcast(exp, v), A)`).

I (somewhat) agree about this point.  But I note that `f..(x)` is actually previously considered (https://github.com/JuliaLang/julia/issues/16285#issuecomment-237320513) but I think it's now discarded as it is already a valid syntax (https://github.com/JuliaLang/julia/issues/19198#issuecomment-388483444; you can see that it's `:(f .. x) == Expr(:call, :.., :f, :x)`).  So, I don't think it's super crazy to use `@..` here as it's already considered before separately.

dlfivefifty:
Ah, we shouldn't use `..` as `a .. b` is the interval `a ≤ x ≤ b` in IntervalSets.jl!

tkf:
OK. Avoid introducing new names makes sense.

mcabbott:
I'm quite keen on writing `sum(@~ A .+ B .* B'; dims=2)`, this was my reason for wanting a macro which made a BroadcastArray, not just a Broadcasted as in the earlier discussion (19198 etc). 

Having to write `sum(AbstractArray(@lazy A .+ B .* B')), dims=2)` seems clumsy, would this gain anything in terms of inference? I don't understand much but `AbstractArray{T,N}` implies eltype must be known I guess. Could the pure Broadcasted object be made efficiently `sum`-able without doing this? 

I presume it's not just me, and that not creating a big array you are about to reduce (over some dimensions) is a big reason you'd want broadcasting.

tkf:
> Could the pure Broadcasted object be made efficiently `sum`-able without doing this?

I believe so.  Actually, using `broadcasted` and `applied` is likely be more efficient for "not so large" arrays as the compiler can eliminate allocations for those object.  I think allocation of the `BroadcastArray` object cannot be eliminated since it invokes the inference API (and hence not type stable).

Efficient `sum` would be implemented via `mapreduce`.  This means that the we need to send a PR to Julia.  But I think this is good since we get better performance in all Julia ecosystem even outside LazyArray.jl.

mcabbott:
OK, so something like adding a method `mapreducedim!(f, op, R::AbstractArray, A::Broadcasted)`, and all the steps between `sum` and that? But one of those needs to create the output array. (Although `sum!(C, @~ A .+ B .* B')` would also be useful.)

For the complete reduction, chethega's `airsum` split off `s = x[1]` first. Does that count as avoiding inference? This stuff may be above my pay grade...

tkf:
By "need inference", what I meant was "need to call inference _API_", i.e., this:

https://github.com/JuliaArrays/LazyArrays.jl/blob/8f2e313fc790fc08dbe6d9cfb517b33ee24c502b/src/lazybroadcasting.jl#L17

It looks like `combine_eltypes` itself can actually be inferred (e.g., try `@code_warntype Broadcast.combine_eltypes(+, ([0.0], [1], [missing, 1im]))`) although I don't know if it fails in more complex ones.  So I probably don't need to worry too much.  But using inference API is something advised to be avoided so I'd say it's safer to not use it.  Another argument is that you can opt-in array version by wrapping the lazy macro with `AbstractArray` but the other way around is impossible.

tkf:
I found a horrible hack to define a macro `@::` :laughing:

```julia
julia> ex = :(macro _(ex); :(lazy.($(esc(ex)))); end);

julia> ex.args[1].args[1] = :(::)
:(::)

julia> eval(ex)
@:: (macro with 1 method)

julia> @macroexpand @:: a .+ b
:(Main.lazy.(a .+ b))
```

nalimilan:
> It looks like `combine_eltypes` itself can actually be inferred (e.g., try `@code_warntype Broadcast.combine_eltypes(+, ([0.0], [1], [missing, 1im]))`) although I don't know if it fails in more complex ones. So I probably don't need to worry too much. But using inference API is something advised to be avoided so I'd say it's safer to not use it. Another argument is that you can opt-in array version by wrapping the lazy macro with `AbstractArray` but the other way around is impossible.

See also https://github.com/JuliaLang/julia/pull/30485.

But indeed it's not recommended to expose information obtained through inference to the user (e.g. via the element type of the array). Since the element type of `broadcast` depends on the actual contents of the input data (e.g. whether there are missing values or not), it cannot be type stable. Inference can only compute an upper bound unless the input type is concrete (neither abstract nor `Union`).

What happens in practice if you don't define the `eltype` of `BroadcastArray`, or always return `Any`? Maybe that would work?

dlfivefifty:
OK, we have the following different cases:

1. Materialized broadcasted. With macro `@.`
2. Lazy broadcasted. Not exposed in Base. 
3. BroadcastArray. No macro has been proposed, but could be `@a.`
4. Materialized applied.
3. Lazy applied. Macro proposed `@~`.
4. ApplyArray. Macro could be `@a~`.

So there's a case for macros for 3,5, and 6.  What about 2 and 4?

mcabbott:
I like the list. The present macro `@~` of this PR is 3. right? And 2. is the `@lazy` from [the discourse thread](https://discourse.julialang.org/t/boolean-short-circuit-fusion/19641/20). Does 4. mean that `Broadcast.materialize` is lazily applied somehow? 

Just to have toy examples somewhere, this is the behaviour with some unions. I don't know how objectionable these are, `BroadcastArray{Union{...` seems preferable to `BroadcastArray{Any,...` surely? 

```julia
julia> ints = [2,3,4];
julia> maybe = Union{Int,Missing}[5,6,7];
julia> f_un(x) = rand() < 0.01 ? missing : x;

julia> ints .+ maybe' # 3×3 Array{Int64,2}

julia> mat = @~ ints .+ maybe' # 3×3 BroadcastArray{Union{Missing, Int64}, ...

julia> collect(@~ int .+ maybe') # 3×3 Array{Union{Missing, Int64},2}; same for copy(), Array()

julia> sum(int .+ maybe'; dims=1) # 1×3 Array{Int64,2}

julia> sum(@~ int .+ maybe'; dims=1) # 1×3 Array{Union{Missing, Int64},2}

julia> f_un.(ints) # 3-element Array{Int64,1}, usually! 

julia> @~ f_un.(ints) # 3-element BroadcastArray{Union{Missing, Int64}
```
I think we should make `collect` call `Broadcast.materialize` here, there should be an easy way to do that. It's possible that `copy` should too, e.g. for lazy transpose it makes a full `Array`. 

Setting `maybe[2] = missing` will change `mat` thus it can't have a narrower type. 
But would it be possible to make `sum(::BroadcastArray, dims=1)` magically (unstably) produce `Array{Int64}`? That ones seems sub-optimal.

mcabbott:
ps. That hack for `@::` also lets you define `@...`, so 1 to 4 on the list are covered! 
```julia
julia> ex.args[1].args[1] = :(...)
:...

julia> @macroexpand @... a .+ b
:((Main.lazy).(a .+ b))
```

dlfivefifty:
> Does 4. mean that `Broadcast.materialize` is lazily applied somehow

In the PR https://github.com/JuliaArrays/LazyArrays.jl/pull/17, for  `DefaultApplyStyle`s, a materialized `applied` is the same as just calling the functions:
```julia
materialize(applied(f, applied(g, x, y)) === f(g(x,y))
````
But there can be other `ApplyStyle`s that materialize in a different way. This is used for `Mul` (which corresponds to `applied(*, ...)`) so that the `materialize` is based on `MemoryLayout`, instead of type, thereby avoiding the issues with `mul!`. I.e.
```julia
materialize(Mul(A,B)) === materialize(applied(*, A, B)) ≠ A*B
```

Another example where this is useful would be matrix-exponentials times vectors. That is,
```julia
materialize(applied(*, applied(exp, A), v))
```
could use the fact that `exp(A)*v` can be done without constructing `exp(A).

tkf:
> What happens in practice if you don't define the `eltype` of `BroadcastArray`, or always return `Any`?

@nalimilan I think the performance would be bad.  For example, `similar` would return an `Array{Any}` which would affect any functions that has to allocate buffer or output collection.  I think having a basic set of API not relying on inference (`broadcasted` and `applied`) and a high-level easy-to-use API based on inference (`BroadcastArray` and `ApplyArray`) is a good balance.

tkf:
> OK, we have the following different cases:

@dlfivefifty I think it's better to have just one macro that composes well.  That is to say, I suggest to implement `@~ expression` that _recursively_ replaces all dot calls with `broadcasted` and all non-dot calls with `applied`.  By recursively I mean that apply the same rule to dot calls _inside_ non-dot calls and vice versa.  That is to say, `@~ (A .* B) * C .+ D` would be lowered to

```julia
x = broadcasted(*, A, B)
y = applied(*, x, C)
broadcasted(+, y, D)
```

Note that this recursive nature is compatible with how `@.` applies the dot to all calls.

With this approach, we cover all the list by composing `@~` and `@.`:

1. Materialized broadcasted: `@. EXPR`
2. Lazy broadcasted: `@~ @. EXPR`
3. BroadcastArray: `AbstractArray(@~ @. EXPR)`
4. Materialized applied: `lhs .= @~ EXPR` or `copy(@~ EXPR)`
5. Lazy applied: `@~ EXPR`
6. ApplyArray: `AbstractArray(@~ EXPR)`.

Note: if you want to reduce the length by 1, you can also write `@~@.`

I know that 3 and 6 are rather long.  However:

* You'd want to have an array when you are at "interface boundary".  In this case, it's better to be explicit if you want a lazy or materialized array.  For example, you better materialize the array before passing it to `*`.

* We don't need a `BroadcastArray` for invoking `mapreduce`/`mapfoldl` if we "fix" them in `Base`.  I think this covers most of the cases for lazy reduction (e.g., `sum`).

tkf:
FYI I started implementing `mapreduce` for `Broadcasted`.  `sum(Broadcast.broadcasted(*, a, b))` is now 5x faster than before and almost as fast as `dot`. https://github.com/JuliaLang/julia/compare/master...tkf:bcreduce

mcabbott:
This plan sounds great, if just a little more ambitious than my PR! Is the idea that `@~` would be part of this package, or would be extended somehow to add 3-6?

tkf:
Yes, my idea for `@~` makes sense only if `@~` lives in LazyArrays.jl since it needs `applied` which is defined in LazyArrays.jl.

dlfivefifty:
There is a case for `applied` to live in its own lightweight package (LazyOperations.jl?) separate from lazy arrays, but I think that's not terribly important right now. 

I'll try to finish the `Applied` PR the next few days.

nalimilan:
> @nalimilan I think the performance would be bad. For example, `similar` would return an `Array{Any}` which would affect any functions that has to allocate buffer or output collection. I think having a basic set of API not relying on inference (`broadcasted` and `applied`) and a high-level easy-to-use API based on inference (`BroadcastArray` and `ApplyArray`) is a good balance.

Right. But maybe that just means it shouldn't support `similar`, as it's not possible to determine the type of the result without relying on inference (which really isn't recommended).

dlfivefifty:
I’m not sure I understand, ` Broadcasted ` supports similar without issue.

nalimilan:
Not AFAICT:
```julia
julia> similar(Base.Broadcast.broadcasted(+, [1]))
ERROR: MethodError: no method matching similar(::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{1},Nothing,typeof(+),Tuple{Array{Int64,1}}})
```

You need to provide the element type:
```julia
julia> similar(Base.Broadcast.broadcasted(+, [1]), Int)
1-element Array{Int64,1}:
 0
```

dlfivefifty:
Right, but one could easily define a `similar(::Broadcasted)`:
```
julia> bc = Broadcast.broadcasted(+, [1]);

julia> similar(bc, Broadcast.combine_eltypes(bc.f, bc.args))
1-element Array{Int64,1}:
 0
```

nalimilan:
`combine_eltypes` depends on inference, so it can return `Any` if inference fails:
```julia
julia> bc = Broadcast.broadcasted(x -> x > 0 ? (x > 1 ? (x > 2 ? nothing : 1) : 1.0) : "a", [1]);

julia> Broadcast.combine_eltypes(bc.f, bc.args)
Any
```

tkf:
> 3\. BroadcastArray: `AbstractArray(@~ @. EXPR)`

Talking to myself: This actually is a type piracy.  So maybe it's better to have a factory function `array` so that `array(@~ @. EXPR)` works.

mcabbott:
> 3. BroadcastArray: `AbstractArray(@~ @. EXPR)`

How does this differ from `BroadcastArray(@~ @. EXPR)`?

> 2\. `@~ @. EXPR`

If `sum(::Broadcast; dims)` ends up in Base, then perhaps the this level 2+4 macro should join it? With the stronger macro of 4,5,6 living here (tell people that `@≈` is two waves?)

tkf:
> How does this differ from `BroadcastArray(@~ @. EXPR)`?

It's nice to be able to write `array(x)` where `x` can be a `broadcasted` or `applied`.

> If `sum(::Broadcast; dims)` ends up in Base, then perhaps the this level 2 macro should join it?

I think it would go in to `Base` as soon as the discussion in https://github.com/JuliaLang/julia/issues/19198 is settled.  I'm trying to speed up it a little bit by making `mapreduce` faster in https://github.com/JuliaLang/julia/pull/31020.

> I also wonder if there is room for confusion between these levels. If `@~ f(A)` is level 5, lazy applied, then does `@~ f(A) .+ B` make a lazy application of `f` in addition to lazy broadcasting?

Since it is _syntactically_ clear that `f(X)` is a call and `f(A) .+ B` is a dot-call (with a call inside its argument), I don't think there is any confusion.  Also, in the end, you just get some `thing` that can be materialized via `lhs .= thing` for `copy(thing)`.  I think it's nice to treat them uniformly.

dlfivefifty:
I really don't like `array` as its too generic and could be confusing. We could do:
```julia
abstract type LazyArray{T,N} <: AbstractArray{T,N} end
struct BroadcastArray{T,N} <: LazyArray{T,N} end
struct ApplyArray{T,N} <: LazyArray{T,N} end
```
then support `LazyArray(@~ @. EXPR)`

tkf:
> I really don't like `array` as its too generic and could be confusing.

This makes sense. `LazyArray(@~ @. EXPR)` sounds good to me.

tkf:
Here is a sample implementation of what I proposed in https://github.com/JuliaArrays/LazyArrays.jl/pull/21#issuecomment-462014552 which can do something like

```julia
julia> @macroexpand @~ a .+ f(b, c)
:((Base.Broadcast.instantiate)((lazy).(a .+ (LazyArrays.applied)(f, b, c))))

julia> @macroexpand @~ a .+ f(b, c .+ d)
:((Base.Broadcast.instantiate)((lazy).(a .+ (LazyArrays.applied)(f, b, (Base.Broadcast.instantiate)((lazy).(c .+ d))))))
```

Code:

```julia
using LazyArrays

lazy(::Any) = throw(ArgumentError("function `lazy` exists only for its effect on broadcasting, see the macro @~"))
struct LazyCast{T}
    value::T
end
Broadcast.broadcasted(::typeof(lazy), x) = LazyCast(x)
Broadcast.materialize(x::LazyCast) = BroadcastArray(x.value)


is_call(ex::Expr) =
    ex.head == :call && !startswith(String(ex.args[1]), ".")

is_dotcall(ex::Expr) =
    ex.head == :. || (ex.head == :call && startswith(String(ex.args[1]), "."))
# e.g., `f.(x, y, z)` or `x .+ y .+ z`

lazy_expr(x) = x
function lazy_expr(ex::Expr)
    if is_dotcall(ex)
        return bc_expr(ex)
    elseif is_call(ex)
        return app_expr(ex)
    else
        # TODO: Maybe better to support `a ? b : c` etc.? But how?
        return ex
    end
end

function bc_expr(ex::Expr)
    @assert is_dotcall(ex)
    return :($(Broadcast.instantiate)($lazy.($(bc_expr_impl(ex)))))
end

bc_expr_impl(x) = x
function bc_expr_impl(ex::Expr)
    # walk down chain of dot calls
    if is_dotcall(ex)
        return Expr(ex.head,
                    lazy_expr(ex.args[1]), # function name (`f`, `.+`, etc.)
                    bc_expr_impl.(ex.args[2:end])...) # arguments
    else
        return lazy_expr(ex)
    end
end

function app_expr(ex::Expr)
    @assert is_call(ex)
    # instantiate?
    return app_expr_impl(ex)
end

app_expr_impl(x) = x
function app_expr_impl(ex::Expr)
    # walk down chain of calls and lazy-ify them
    if is_call(ex)
        return :($applied($(app_expr_impl.(ex.args)...)))
    else
        return lazy_expr(ex)
    end
end

macro ~(ex)
    # checkex(ex)
    esc(lazy_expr(ex))
end
```

dlfivefifty:
Let me know when this is ready to commit.

PR #31, opened 2019-04-29
Support creating Applied/Broadcasted mixes via @~

tkf:
This PR implements my suggestion https://github.com/JuliaArrays/LazyArrays.jl/pull/21#issuecomment-464388713 so that we can do

```julia
Y .= @~ α .* (A * x) .+ β .* c
```

The main changes are:

1. `@~` converts normal calls to `applied`.
2. An abstract type `LazyArray` is added as suggested in https://github.com/JuliaArrays/LazyArrays.jl/pull/21#issuecomment-462656843.  It is the supertype of `BroadcastArray` and `ApplyArray`.
2. `@~` does not wrap `Broadcasted` in `BroadcastArray` anymore.  Instead, users have to call `LazyArray` explicitly.
3. `@~` calls `instantiate` on `Broadcasted`s (this is the same as https://github.com/JuliaLang/julia/pull/31020)

@mcabbott probably does not like that we need to write `sum(LazyArray(@~ ...))` to get the speedup.  However, that's only until https://github.com/JuliaLang/julia/pull/31020 is merged.

I think it's nice to have an API that does not uses inference API so that `y .= @~ ...` can avoid inference overhead. It may be useful, e.g., inside a tight loop.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,15 +82,15 @@ julia> B == A .+ 2
 true
 ```
 Such arrays can also be created using the macro `@~` which acts on ordinary 
-broadcasting expressions: 
+broadcasting expressions combined with `LazyArray`:
 ```julia
 julia> C = rand(1000)';
 
-julia> D = @~ exp.(C)
+julia> D = LazyArray(@~ exp.(C))
 
-julia> E = @~ @. 2 + log(C)
+julia> E = LazyArray(@~ @. 2 + log(C))
 
-julia> @btime sum(@~ C .* C'; dims=1) # without `@~`, 1.438 ms (5 allocations: 7.64 MiB) 
+julia> @btime sum(LazyArray(@~ C .* C'); dims=1) # without `@~`, 1.438 ms (5 allocations: 7.64 MiB)
   74.425 μs (7 allocations: 8.08 KiB)
 ```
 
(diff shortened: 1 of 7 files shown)


dlfivefifty:
What do you think about using `Applied` instead of hacking `Broadcasted` for the BLAS MulAdd operations? That is,
```julia
Y .= @~ α .* (A * x) .+ β .* c
```
would become
```julia
Y .= @~ α * A * x + β * c
```
which is code for
```julia
Y .= applied(+, applied(*, α, A, x), applied(*, β, c))
```

dlfivefifty:
Using `ApplyStyle` we could also have a much more robust way of rearranging complicated `MulAdd` operations to reduce to `MulAdd`.

tkf:
`@~ α * A * x + β * c` sounds good to me.  It looks cleaner.

In a long run, I think it would be nice if both `@~ α .* (A * x) .+ β .* c` and `@~ α * A * x + β * c` can be "normalized" to a same representation via `instantiate` mechanism.  This way, users don't have to know when to use, e.g., `+` or `.+` to invoke performance specializations.

dlfivefifty:
Yes, I suppose both could work by an intelligent use of `ApplyStyle` and `BroadcastStyle`.

In the meantime, I'll go ahead and merge this PR.

Issue #33, opened 2019-05-01
Replace special types with ApplyArray


dlfivefifty:
All of these types are special cases of `ApplyArray`:

- [ ] `Kron` (`ApplyArray(kron, a, b)`)
- [ ] `Diff`
- [ ] `Cumsum`
- [ ] `Vcat`
- [ ] `Hcat`

Removing them as special cases will lead to more unified code

dlfivefifty:
I think this has largely been done.

PR #34, opened 2019-05-06
Setup Documenter

tkf:
This PR sets up Documenter. I leave out `deploydocs` setup so you still need to do the usual Travis setup.

I'm sending this PR because it would motivate me to write docstring next time I send a feature PR :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,3 +20,14 @@ after_success:
 #script:
 #  - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
 #  - julia -e 'Pkg.clone(pwd()); Pkg.build("MyPackage"); Pkg.test("MyPackage"; coverage=true)'
+
+jobs:
+  include:
+    - stage: Documentation
+      julia: 1.1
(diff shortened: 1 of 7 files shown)


dlfivefifty:
Unfortunately I don't have time to work on docs right now, but I've added you as a collaborator so feel free to merge whatever you feel doesn't require a review.

Note that it might be worth waiting to the code settles down a bit more. For example:

https://github.com/JuliaArrays/LazyArrays.jl/issues/35
https://github.com/JuliaArrays/LazyArrays.jl/issues/33

and also I'm thinking that `BroadcastArray` and `ApplyArray` should not wrap a `Broadcasted` and `Applied` instance, instead have the same fields (and so make dispatch a bit simpler).

tkf:
> I've added you as a collaborator

Thanks!

I don't think that's enough access to setup Travis/Documenter. It's not urgent at all, but can you do it when you have time #36?

> `BroadcastArray` and `ApplyArray` should not wrap a `Broadcasted` and `Applied` instance, instead have the same fields

I think it makes sense.

PR #41, opened 2019-06-21
Add Interlace

dlfivefifty:


DanielVandH:
Was there any particular reason this stalled if you remember @dlfivefifty? I can see a version of this for infinite arrays here

https://github.com/JuliaApproximation/AlgebraicCurveOrthogonalPolynomials.jl/blob/5bf062d232194069b4d41a35a35b86950c3ed533/src/twoband.jl#L83-L93

but a `LazyInterlace` could be more broadly useful here, right?

dlfivefifty:
The whole interlace business is just in a state of mess because I wasn't entirely sure the design.

The first question is: is there a need for `Interlace`? Here are two synonyms:
```julia
julia> a = [1,2,3]; b = [4,5,6]; LazyBandedMatrices.BlockVec(Hcat(a,b)')
blockvec(vcat(1×3 adjoint(::Vector{Int64}) with eltype Int64, 1×3 adjoint(::Vector{Int64}) with eltype Int64)) with indices BlockedOneTo(2:2:6):
 1
 4
 ─
 2
 5
 ─
 3
 6

julia> BlockBroadcastArray(vcat, unitblocks(a), unitblocks(b))
3-blocked 6-element BlockBroadcastArray{Int64, 1, typeof(vcat), Tuple{BlockedVector{Int64, Vector{Int64}, Tuple{BlockedOneTo{Int64, Base.OneTo{Int64}}}}, BlockedVector{Int64, Vector{Int64}, Tuple{BlockedOneTo{Int64, Base.OneTo{Int64}}}}}}:
 1
 4
 ─
 2
 5
 ─
 3
 6
```

The second question: how to unify the different version of concatentation in ContinuumArrays.jl? See:

1. https://github.com/JuliaApproximation/ContinuumArrays.jl/blob/master/src/bases/basisconcat.jl
2. https://github.com/JuliaApproximation/ClassicalOrthogonalPolynomials.jl/blob/main/src/interlace.jl

The types defined in the latter are more thoroughly developed but I don't remember why its in ClassicalOrthogonalPolynomials.jl. It should probably be moved to ContinuumArrays.jl and unified with the types in basisconcat.jl.

We can probably discuss this in more detail at the next meeting.

Issue #35, opened 2019-05-06
`ApplyStyle` should not need instantiation


dlfivefifty:
It feels wrong that `BroadcastStyle` is based completely on type, not on instantiation, while `ApplyStyle` allows instantiation. This should be made consistent. 

One issue is the conflation of `MemoryLayout` and `ApplyStyle`. I think since most operators pipe through `MulAdd` this can be redesigned so that the corresponding apply style is a flag saying `MulAddStyle` and the use of `MemoryLayout` is confined to `MulAdd`.

dlfivefifty:
Done.

PR #46, opened 2019-08-11
Simplify ApplyStyle

dlfivefifty:
The goal here is to move `Applied` closer to `Broadcasted` in implementation. For example, `ApplyStyle` will only depend on types, and information about `MemoryLayout` will be left to specialised types a la the existing `MulAdd`.

This should hopefully also reduce compile time.


dlfivefifty:
@tkf This is progressing quite nicely so feel free to take a look and see whether you have any big comments before I wrap it up.

One of the big changes (which I really like) is that `MulAdd` shorthands become, e.g.:
```julia
c .= @~ A*x
c .= @~ A*x + c
c .= @~ α*A*x + β*c
```
This replaces `c .= α .* Mul(A,x) .+ β .* c`. It's a more extendable design as it's now based on a special `ApplyStyle` so in principle this could be taken further (e.g., `c .= @~ c + A*x`) though I won't take it further yet. 

One negative is that it is _not_ possible to do any of these yet:
```julia
c .= @~ A'*x # the transpose is made lazy which ruins the style
c .= @~ 2A*x # the `2A` is grouped together which isn't supported yet
c .= @~ one(T) * A * x # the `one(T)` is made lazy which ruins the style
```
I'm not sure how important these are to fix right now.

src/lazyapplying.jl
```diff
@@ -3,19 +3,29 @@
...
+Applied{Style}(f::F, args::Args) where {Style,F,Args<:Tuple} = 
+    Applied{Style,Core.Typeof(f),Args}(f, args)
+
+_typesof() = ()
+_typesof(a, b...) = tuple(typeof(a), _typesof(b...)...)
+_typesof(a, b) = tuple(typeof(a), typeof(b))
+_typesof(a, b, c) = tuple(typeof(a), typeof(b), typeof(c))
+combine_apply_style(f, a...) = ApplyStyle(f, _typesof(a...)...)
+combine_apply_style(f, a, b) = ApplyStyle(f, typeof(a), typeof(b))
+combine_apply_style(f, a, b, c) = ApplyStyle(f, typeof(a), typeof(b), typeof(c))
```
tkf: I wonder if we need `combine_apply_style`.  As a reference, here is how `combine_styles` is used:

```julia
@inline function broadcasted(f, args...)
    args′ = map(broadcastable, args)
    broadcasted(combine_styles(args′...), f, args′...)
end
```

In broadcasting you need to combine styles of all arguments because they have to negotiate each other to find out the best way to do `getindex`.  But in `Applied`, you don't have this problem because function "sees" the entire arguments, not just their elements.

tkf:
Thanks, `c .= @~ α*A*x + β*c` etc. are beautiful!

> The goal here is to move `Applied` closer to `Broadcasted` in implementation.

To make it even closer, how about adding `LazyArrays.instantiate` as well?  The motivation is to make `@~` a canonical surface syntax of lazy APIs in Julia packages.  For example, when I export an eager function `f` from `MyModule`, I may want implement an lazy version of `f` which require temporary arrays.  If there is `LazyArrays.instantiate`, I can implement

```julia
LazyArrays.instantiate(app::Applied{<:Any, typeof(f)}) = FSolver(app.args...)
Baes.copyto!(dest::AbstractArray, solver::FSolver) = solve!(dest, solver)
```

in `MyModule` so that I can reuse the solver via

```julia
input = ...
lazy_sol = @~ f(input)
solutions = [
    begin
        input .= x
        copy(lazy_sol)
    end
    for x in ...
]
```

See also: https://discourse.julialang.org/t/lazy-and-eager-api-design/22398/6

> the transpose is made lazy which ruins the style

Can `instantiate` help?  I imagine something like

```julia
struct Eager <: ApplyStyle end
ApplyStyle(::typeof(adjoint), _) = Eager()

instantiate(app::Applied{Eager}) = app.f(instantiate.(f.args...))
```

where the default `instantiate` is

```julia
instantiate(x) = x
instantiate(app::Applied{S}) where {S} = Applied{S}(app.f, instantiate.(app.args)...)
```

src/lazyapplying.jl
dlfivefifty: I'm really just using it as a way to change from calling on a value to a type. Though it hints at a possible future refinement where combined arguments are handled via `ApplyStyle(f, ::ApplyStyle, ::ApplyStyle...)`. 

I'll leave it for now as it affects whether BLAS calls are not allocating, and so it takes care to redesign it.

dlfivefifty:
Also now resolves

https://github.com/JuliaArrays/LazyArrays.jl/issues/35

dlfivefifty:
This failure is particularly annoying:
```julia
@~ (2.0+0.0im)*A*x + (3.0+0.0im)*y
```
The issue is it tries to make `2.0+0.0im` lazy....

I'm not sure how to handle this yet.

dlfivefifty:
@tkf Do you object if I drop support for Julia v0.7? The test is broken due to `Ref` not working the same. I can work around it by using `(A,)` instead but I'm not sure it's worth bothering to maintain backward compatibility.

tkf:
> Do you object if I drop support for Julia v0.7?

Not at all!

> The issue is it tries to make `2.0+0.0im` lazy....

Maybe `ApplyStyle(::typeof(+), ::Vararg{Type{<:Number}}) = Eager()`?

dlfivefifty:
Or just overload `applied(+,::Number,::Number)` a la
```julia
julia> Base.broadcasted(+,1:3,2)
3:5
```

tkf:
> Or just overload `applied(+,::Number,::Number)`

Yes, it also makes sense.  I just thought it might be better to make it possible to construct a call tree for _any_ functions and arguments.  For example, maybe some people want to do some algebraic simplification or something.  But maybe it's better to not include this kind of generality until someone asks for it.

dlfivefifty:
It is possible by calling `Applied` directly.

tkf:
Good point. But that's not as handy as `@~`, right?

dlfivefifty:
Right. But it doesn’t make sense to make things handy for fictional users when they can use the non-handy version if they ever become non-fictional. Especially since getting ` Eager ` to work with ` MulAdd ` is not trivial, and I’m not eager (🤣) to do it.

tkf:
> But it doesn’t make sense to make things handy for fictional users

I agree.

tkf:
...having said that, `instantiate` phase is useful for at least one non-fictional user (= me).  For example, I want to convert

```julia
rhs = @~ α*A*f.(x .+ y) + β*c
```

to something equivalent to

```julia
rhs = @~ α*A*(tmp .= f.(x .+ y)) + β*c
```

with a properly allocated `tmp` array in the `instantiate` phase so that `c .= rhs` becomes non-allocating.

Of course, I can implement my own function `optimize` and do `rhs′ = optimize(rhs)` so it is not a hard requirement.

dlfivefifty:
I’ve already added ` instantiate ` ...

tkf:
Ah, sorry, I missed it.

tkf:
#47 caught an error in

https://github.com/JuliaArrays/LazyArrays.jl/blob/df8f179c8b2e81fae471adf6f2ba2cf392541b23/src/linalg/mul.jl#L9-L12

Should it support `Broadcast.instantiate(applied(*, 1, ones(2, 2)))`?

PR #47, opened 2019-08-14
Call instantiate on Applied in @~ macro

tkf:
This is a PR against #46.

Now that we have `instantiate(::Applied) `, I think the best strategy to call it in `@~` is to do this in the "root" node and let it propagate through recursion.


dlfivefifty:
I'll merge #46 first and then we can resolve this

Issue #48, opened 2019-08-15
Block product


antoine-levitt:
The following is slow
```
using BenchmarkTools
using LazyArrays

X = randn(1000,100)
Yh = hcat(X,X)
YH = Hcat(X,X)
@btime $Yh'*$Yh
@btime $YH'*$YH
```
Am I doing something wrong? I was hoping this package would cleverly dispatch to BLAS here.

antoine-levitt:
Hm, I guess the problem is that it's tricky to do when the subarray boundaries don't align. I'd be willing to try my hand on a PR to implement this in the case where the boundaries are aligned (I need this functionality pretty badly), but is it OK to error when they are not? I'm guessing the users of this package are here for performance, so it seems acceptable. If not, I'm not sure how to call the fallback from the specialization...

dlfivefifty:
Yes it's fine to throw an error. Note that `MemoryLayout` for `Vcat` is already implemented:
```julia
julia> LazyArrays.MemoryLayout(typeof(YH'))
LazyArrays.VcatLayout{Tuple{LazyArrays.DenseRowMajor,LazyArrays.DenseRowMajor}}()
```
Thus the easiest way is to add an `HcatLayout` and overload `materialize!`, something like:
```julia
function materialize!(M::MulAdd{<:VcatLayout,<:HcatLayout})
   α,A,B,β,C =  M.α,M.A,M.B,M.β,M.C
   T = eltype(C)
   _fill_lmul!(β,C) # this is temporary until strong β = false is supported
   for (a,b) in zip(A.arrays,B.arrays)
       materialize!(MulAdd(α,a,b,one(T),C))
   end
   C
end
```

antoine-levitt:
I'm confused, this is for the lazy Mul, right? It won't do anything for my MWE above?

dlfivefifty:
As long as we add `@lazymul Vcat` it will automatically lower to `materialize!(::MulAdd)`:
```julia
julia> LazyArrays.@lazymul Vcat

julia> @which YH'YH # calls materialize!(Mul(YH',YH)), which defaults to MulAdd
*(A::Vcat, B::AbstractArray{T,2} where T, C...) in Main at /Users/solver/Projects/LazyArrays.jl/src/linalg/lazymul.jl:18
```
The "power" of the lazy memory layout approach is it avoids ambiguity issues.

antoine-levitt:
[...]

PR #50, opened 2019-08-16
Add Hcat * Vcat special case

dlfivefifty:
Starts to address https://github.com/JuliaArrays/LazyArrays.jl/issues/48

@antoine-levitt This should give an idea on what to do. At this point just overloading
```julia
materialize!(::MatMulMatAdd{<:VcatLayout,<:HcatLayout})
```
should suffice.


antoine-levitt:
Thanks! I was just looking at this.

antoine-levitt:
I assume we also need something to support products of cats and regular matrices? Or is that magically covered?

dlfivefifty:
It will fall back to the slow default, but something like the following should be fast for Hcat * Strided (e.g. Hcat * Vector):
```julia
function materialize!(M::MatMulVecAdd{<:HcatLayout})
    α,A,B,β,C =  M.α,M.A,M.B,M.β,M.C
    T = eltype(C)
    _fill_lmul!(β,C) # this is temporary until strong β = false is supported
    ind = 0
    for a in A.arrays
        jr = ind+1:ind+size(a,2)
        b = view(B, jr, :) 
        materialize!(MulAdd(α,a,b,one(T),C))
        ind = last(jr)
    end
    C
 end
```

antoine-levitt:
OK, thanks! Will try to do the 3^2 combinations with Hcat,Vcat and dense.

dlfivefifty:
I think you only need 4, not 3^2: ` Hcat * Vcat `, ` Vcat * Hcat `, ` Vcat * any `, ` Hcat * any `. This is because I don’t think ` Vcat * Vcat ` or ` Hcat * Hcat ` are that useful....

(Ok each will need to be done three times, once for matrix * matrix, matrix * vector and vector * matrix...)

PR #56, opened 2019-09-17
Fix copyto! for multi-argument ApplyArrayBroadcastStyle

### LINKED ISSUE #54: Wrong result when Applied is inside Broadcasted
There seems to be an issue when you put an Applied inside Broadcasted:

```julia
julia> A = ones(1, 1); x = [1];

julia> app = @~ A * x
       y = @~ exp.(app)
Base.Broadcast.Broadcasted{LazyArrays.ApplyArrayBroadcastStyle{1}}(exp, ([1.0]⋆[1],))

julia> copy(y)
1-element Array{Float64,1}:
 1.0
```

but it should be

```julia
julia> exp.(copy(app))
1-element Array{Float64,1}:
 2.718281828459045
```

It looks like `bc.f` (= `exp`) is not applied here:

https://github.com/JuliaArrays/LazyArrays.jl/blob/71e51a0c33b465054d87c01b75c7af28b8925e33/src/lazyapplying.jl#L86-L89

Should it do something like this?

```julia
@inline function copyto!(dest::AbstractArray, bc::Broadcasted{ApplyArrayBroadcastStyle{N}}) where N
    @assert length(bc.args) == 1
    copyto!(dest, first(bc.args))
    map!(bc.f, dest)
    return dest
end
```

tkf:
fix #54.

### DIFF
--- a/src/lazyapplying.jl
+++ b/src/lazyapplying.jl
@@ -84,9 +84,29 @@ similar(bc::Broadcasted{ApplyArrayBroadcastStyle{N}}, ::Type{ElType}) where {N,E
     copyto!(dest, first(bc.args))
 end
 @inline function copyto!(dest::AbstractArray, bc::Broadcasted{ApplyArrayBroadcastStyle{N}}) where N 
-    @assert length(bc.args) == 1
-    copyto!(dest, first(bc.args))    
+    if length(bc.args) == 1
+        copyto!(dest, first(bc.args))
+        if bc.f !== identity
+            dest .= bc.f.(dest)
+        end
+    else
+        bc′ = mapbc(bc) do x
+            if x isa Applied
+                materialize(x)
+            else
+                x
+            end
+        end
+        materialize!(dest, bc′)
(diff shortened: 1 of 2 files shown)


src/lazyapplying.jl
```diff
@@ -84,10 +84,30 @@ similar(bc::Broadcasted{ApplyArrayBroadcastStyle{N}}, ::Type{ElType}) where {N,E
     copyto!(dest, first(bc.args))
 end
 @inline function copyto!(dest::AbstractArray, bc::Broadcasted{ApplyArrayBroadcastStyle{N}}) where N 
-    @assert length(bc.args) == 1
-    copyto!(dest, first(bc.args))    
+    if length(bc.args) == 1
+        copyto!(dest, first(bc.args))
+        if bc.f !== identity
+            dest .= bc.f.(dest)
```
tkf: This is a bit dangerous definition considering input and output types of `bc.f` can be different.

PR #61, opened 2019-09-18
Improve Lmul

dlfivefifty:


dlfivefifty:
@tkf if you are able to review this while I get the coverage up it would be great.  This PR

1. Adds `Rmul`
2. Standardised `Lmul`
3. Simplifies `ApplyLayout` and `BroadcastLayout`. The idea being that if you want more information you should override `applylayout` or `broadcastlayout` to return a custom layout. An example is `PaddedLayout` for vectors padded with zeros.
4. Minor changes to `LazyLayout` for downstream packages

src/linalg/lmul.jl
```diff
@@ -0,0 +1,233 @@
...
+# equivalent to rescaling
+function materialize!(M::Lmul{<:DiagonalLayout{<:AbstractFillLayout}})
+    M.B .= getindex_value(M.A.diag) .* M.B
+    M.B
+end
+
+copy(M::Lmul{<:DiagonalLayout,<:DiagonalLayout}) = Diagonal(M.A.diag .* M.B.diag)
+copy(M::Lmul{<:DiagonalLayout}) = M.A.diag .* M.B
+copy(M::Lmul{<:DiagonalLayout{<:AbstractFillLayout}}) = getindex_value(M.A.diag) .* M.B
+copy(M::Lmul{<:DiagonalLayout{<:AbstractFillLayout},<:DiagonalLayout}) = Diagonal(getindex_value(M.A.diag) .* M.B.diag)
```
tkf: Why not define `materialize!` and let the default `copy(M::Lmul)` do the job? (Same question for `Rmul`)

dlfivefifty: Because in InfiniteArrays.jl/InfiniteLinearAlgebra.jl `M.A` and `M.B` are ∞-dimensional, and these result in a lazy `BroadcastArray`

dlfivefifty: And the special case for `FillLayout` is because if they are ∞-dimensional banded matrices scalar multiplication lowers down. So we can do:
```julia
A = _BandedMatrix((1:∞)', ∞, -1, 1)
2.0*A === _BandedMatrix((2.0:2.0:∞)', ∞, -1, 1)
```
which is useful for fast building of spectral methods.

src/lazyapplying.jl
```diff
@@ -214,10 +217,14 @@ result_mul_style(::LazyArrayApplyStyle, _) = LazyArrayApplyStyle()
...
+struct  ApplyLayout{F} <: MemoryLayout end
+
+applylayout(::F, args...) where F = ApplyLayout{F}()
 
-MemoryLayout(M::Type{Applied{Style,F,Args}}) where {Style,F,Args} = ApplyLayout{F,tuple_type_memorylayouts(Args)}()
-MemoryLayout(M::Type{ApplyArray{T,N,F,Args}}) where {T,N,F,Args} = ApplyLayout{F,tuple_type_memorylayouts(Args)}()
+MemoryLayout(::Type{Applied{Style,F,Args}}) where {Style,F,Args} = 
+    applylayout(F.instance,tuple_type_memorylayouts(Args).parameters...)
+MemoryLayout(::Type{ApplyArray{T,N,F,Args}}) where {T,N,F,Args} = 
+    applylayout(F.instance,tuple_type_memorylayouts(Args).parameters...)
```
tkf: I guess `F.instance` would break for general callables? Though I can see that typing `::Type{typeof(+)}` etc. would be annoying. But maybe using `F` instead of `F.instance` with a helper shortcut function `TypeOf(f) = Type{typeof(f)}` is OK?

dlfivefifty: Yes I agree this is less than ideal, and you are right that the motivation was just to make it less "annoying". It's probably also what's breaking type inference on 1.0.

I think the helper function is overkill since these are one-off implementations. 

I'll sleep on it... my only hesitation on reverting to the less clever version is it feels "harder" on the compiler, in some sense that's too early optimisation but the compiler seems to be quite delicate in these situations. (Though it sounds like its going to improve shortly.)

dlfivefifty: Thinking it more, there’s plenty of types (like ` Fun ` ) that represent functions that cannot be instantiated. So I’ll change this to be less clever

dlfivefifty: Thanks for making me look at this, as I also realised in the process the setup with `tuple_type_memorylayouts` was insane: it was creating a `Type{<:Tuple}` from a tuple, and then re-instantiating to get back the original tuple! It's now fixed to just pass the tuple (and hence the code isinferrable in Julia v1.0)

dlfivefifty:
I'll merge and tag once tests pass

PR #73, opened 2019-10-15
Support views of `ApplyArray(*, ...)` 

dlfivefifty:


dlfivefifty:
@jkf this now allocates due to use of views for some cases it didn't previously on `copyto!` with `Vcat`. Is this a concern for you? I can re-add the special case code (which called `setindex!` instead of `copyto!`) if needed.

Issue #65, opened 2019-09-27
Performance improvement of Kron times Vector/Matrix


martka:
Up to now I was using v0.9.1 of LazyArrays and implemented an improved version of multiplying a Kron with a Vector/Matrix. After updating to v0.12.0 my code is not working 
anymore.

I'm writing a program for statistical analysis of data that makes heavy use of big matrices.
The memory requirement of the matrices that I'm using can easily go up to 100TB, way too big to store them directly in memory. Luckily my matrices are kronecker products of smaller matrices so i am using LazyArrays.Kron to represent them.

In my program i often need to multiply a Kron with a Vector or Matrix. Up to now i was using version 0.9.1 of LazyArrays and the way the multiplication is carried out there is not feasible for me (IIRC generic_matmatmul gets called and each element of the Kron is visited individually).

Therefore, I implemented my own version of `(*)(::Kron,::Vector)` or actually `mul!(::Vector, ::Kron, ::Vector)`. 
In essence it uses the property `Kron(A,B)*vec(X)=vec(B*X*A^T)` ([https://en.wikipedia.org/wiki/Kronecker_product#Matrix_equations](url)) to avoid creating a dense matrix and iterating through all elements of the matrix. If a matrix should be multiplied it does so by multiplying for each column using views of the columns to avoid memory allocation. Higher order kronecker products are handled recursively `Kron(A,B,C)*vec(X)=Kron(B,C)*(X*A^T)`. The actual implementation is of course a bit more complicated, especially for higher order kronecker products where chains of reshape/view occur (the BLAS functions don't like chains of reshape/view).
I was able to achieve a significant performance improvement w.r.t. the 0.9.1 version of LazyArrays and also the straigh-forward Julia implementation using `kron`(not for the huge matrices of course).
```
A=randn(ComplexF64,10,10);
B=randn(ComplexF64,10,10);
C=kron(A,B);
CK=Kron(A,B);
x=randn(size(CK,1));
C*x;CK*x; # warmup
@btime $CK*$x; #  6.861 μs (10 allocations: 6.88 KiB)
@btime $C*$x; #   21.639 μs (1 allocation: 3.25 KiB)
```
Please note that the benchmark for `kron` does not take into account the time it takes to compute `kron(A,B)` and still it performs worse. For higher order kronecker products the performance gain differs but is there and allows me to work with big matrices at all.

Without any changes to v0.9.1 i got this result
```
[...]

PR #78, opened 2019-11-03
Optimizing determinants and traces of Kron

emerali:
See #65 

I've taken the liberty to provide optimized determinant and trace computations of Kronecker products as these operations probably don't need to be done lazily.

I haven't yet implemented Kron * Vector/Matrix multiplication as I need to familiarize myself with the code base a bit more before I can do that, and it also seems like there may be blockage due to #77.

### DIFF
--- a/src/lazyoperations.jl
+++ b/src/lazyoperations.jl
@@ -23,6 +23,43 @@ axes(a::Kron{<:Any,1}) = (OneTo(size(a,1)),)
 axes(a::Kron{<:Any,2}) = (OneTo(size(a,1)), OneTo(size(a,2)))
 axes(a::Kron{<:Any,N}) where {N} = (@_inline_meta; ntuple(M -> OneTo(size(a, M)), Val(N)))
 
+
+function det(K::Kron{<:Any, 2})
+    (size(K, 1) == size(K, 2)) || throw(DimensionMismatch("matrix is not square: dimensions are $(size(K))"))
+
+    d = 1.
+    s = size(K, 1)
+
+    not_square = []
+    for A in K.args
+        if size(A, 1) == size(A, 2)
+            dA = det(A)
+            if iszero(dA)
+                return dA
+            end
+            d *= dA^(s ÷ size(A, 1))
+        else
+            # The Kronecker Product of rectangular matrices, if it is square, will
+            # have determinant zero. This can be shown by using the fact that
+            # rank(A ⊗ B) = rank(A)rank(B) and showing that this is strictly less
+            # than the number of rows in the resulting Kronecker matrix. Hence,
+            # since A ⊗ B does not have full rank, its determinant must be zero.
+            return zero(d)
+        end
+    end
+    return d
+end
+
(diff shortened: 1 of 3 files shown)


dlfivefifty:
Great! Can you add tests?

That PR shouldn’t impact things to much, it’s just moving code out to a separate package.

emerali:
Tests have been added for det and tr. Sorry about the extra commits, had to fix a merge conflict to get the CI to run, we can squash them down later. 

I'll be busy for the better part of this week though, so likely won't be able to work much on this till the weekend.

src/lazyoperations.jl
```diff
@@ -23,6 +23,43 @@ axes(a::Kron{<:Any,1}) = (OneTo(size(a,1)),)
...
 axes(a::Kron{<:Any,N}) where {N} = (@_inline_meta; ntuple(M -> OneTo(size(a, M)), Val(N)))
 
+
+function det(K::Kron{<:Any, 2})
+    (size(K, 1) == size(K, 2)) || throw(DimensionMismatch("matrix is not square: dimensions are $(size(K))"))
+
+    d = 1.
+    s = size(K, 1)
+
+    not_square = []
```
dlfivefifty: Where is not_square used?

test/runtests.jl
```diff
@@ -36,14 +36,33 @@ include("broadcasttests.jl")
...
-    @test [K[k,j] for k=1:size(K,1), j=1:size(K,2)] == Array(Kron(A,B)) == kron(A,B)
-    K = Kron(A,B')
-    @test [K[k,j] for k=1:size(K,1), j=1:size(K,2)] == Array(Kron(A,B')) == kron(A,B')
-    K = Kron(A',B)
-    @test [K[k,j] for k=1:size(K,1), j=1:size(K,2)] == Array(Kron(A',B)) == kron(A',B)
-    K = Kron(A',B')
-    @test [K[k,j] for k=1:size(K,1), j=1:size(K,2)] == Array(Kron(A',B')) == kron(A',B')
+    K, k = Kron(A,B), kron(A,B)
+    @test [K[k,j] for k=1:size(K,1), j=1:size(K,2)] == Array(Kron(A,B)) == k
+    @test det(K) == 0  # kronecker of rectangular factors
```
dlfivefifty: Can we change this to `det(K) == det(k) == 0`? Just to double check that the argument for rectangular factors giving zero determinant is correct 😉

src/lazyoperations.jl
emerali: oops, that was part of an old implementation that I forgot to remove

test/runtests.jl
emerali: Unfortunately couldn't one-line this, as it doesn't seem possible to use `≈` when one side is exactly zero, since `≈` performs relative comparison IIRC.

dlfivefifty:
Remove the WIP when you are happy for this to be merged

emerali:
bump

PR #79, opened 2019-11-04
Allow use with StaticArrays 0.12

c42f:
Tests seem to pass locally for this; API deprecations were very minor (semver seems to be a blunt weapon for dealing with "technically breaking" but not actually very breaking changes...)

XRef https://discourse.julialang.org/t/staticarrays-0-12-0-new-static-array-literal-syntax/30645/3

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -11,7 +11,7 @@ StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 [compat]
 FillArrays = "0.7,0.8"
 MacroTools = "0.4.5,0.5"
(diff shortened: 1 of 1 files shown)


dlfivefifty:
Yes I wish there was a way to automate these updates...

fredrikekre:
> Yes I wish there was a way to automate these updates...

https://github.com/bcbi/CompatHelper.jl

dlfivefifty:
OK I think I anded CompatHelper.jl, I guess we'll see what happens next time there's a tag

PR #86, opened 2019-11-29
Fix for copy #85

### LINKED ISSUE #85: copy of LazyArrays
Long story short, the copy mechanism of apply arrays seems to cause issues in some problems, suggested alternative below. 

context for this issue:  https://github.com/JuliaData/Tables.jl/issues/122 in particular reproducing MWE  https://github.com/JuliaData/Tables.jl/issues/122#issuecomment-559566278

The problem identified by  @nalimilan is:

> AFAICT, the bug is that `copy(A::ApplyArray) = ApplyArray(A.f, map(copy,A.args)...)` when it should be `copy(A::ApplyArray{T,N}) where {T,N} = ApplyArray{T,N}(A.f, map(copy,A.args)...)`. There may be other places with the same problem.

thoughts?

tlienart:
I'm not familiar with the code, please let me know if it should be changed elsewhere for instance in 

```julia
copy(A::Applied{LazyArrayApplyStyle}) = ApplyArray(A)
```

cheers

closes #85

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "LazyArrays"
 uuid = "5078a376-72f3-5289-bfd5-ec5146d43c02"
-version = "0.14.9"
+version = "0.14.10"
 
 
 [deps]
--- a/src/lazyapplying.jl
+++ b/src/lazyapplying.jl
@@ -184,7 +184,7 @@ AbstractArray{T,N}(A::ApplyArray{<:Any,N}) where {T,N} = ApplyArray{T,N}(A.f, ma
 
 @inline axes(A::ApplyArray) = axes(Applied(A))
 @inline size(A::ApplyArray) = map(length, axes(A))
-@inline copy(A::ApplyArray) = ApplyArray(A.f, map(copy,A.args)...)
+@inline copy(A::ApplyArray{T,N}) where {T,N} = ApplyArray{T,N}(A.f, map(copy,A.args)...)
 
 
 struct LazyArrayApplyStyle <: AbstractArrayApplyStyle end
--- a/test/applytests.jl
+++ b/test/applytests.jl
@@ -40,3 +40,12 @@ end
     @test M isa ApplyArray
     @test M == exp(A)
 end
+
+@testset "copy (#85)" begin
+    Base.size(A::Applied) = (length(A.args[1]),)
+    Base.eltype(A::Applied)  = eltype(A.args[1])
+    v = ApplyVector(vec, ones(Int, 2, 2))
+    vc = copy(float.(v))
+    ve = convert(Vector, vc)
+    @test eltype(ve) == Float64
+end


dlfivefifty:
Can you add a test? This is the simplest example I could think of:
```julia
julia> Base.size(A::Applied{<:Any,typeof(vec)}) = (length(A.args[1]),)

julia> Base.eltype(A::Applied{<:Any,typeof(vec)}) = eltype(A.args[1])

julia> ApplyVector(vec,randn(2,2))
4-element ApplyArray{Float64,1,typeof(vec),Tuple{Array{Float64,2}}}:
  0.6256960993168339
 -0.3914615678807611
  1.300003334763862 
  0.6678915282817223

julia> copy(ApplyVector(vec,randn(2,2)))
ERROR: MethodError: no method matching ndims(::Applied{LazyArrays.DefaultArrayApplyStyle,typeof(vec),Tuple{Array{Float64,2}}})
Closest candidates are:
  ndims(::Base.Generator) at generator.jl:53
  ndims(::Number) at number.jl:67
  ndims(::Type{#s66} where #s66<:Number) at number.jl:68
  ...
Stacktrace:
 [1] ApplyArray{Float64,N,F,Args} where Args<:Tuple where F where N(::Applied{LazyArrays.DefaultArrayApplyStyle,typeof(vec),Tuple{Array{Float64,2}}}) at /Users/sheehanolver/.julia/packages/LazyArrays/32TDf/src/lazyapplying.jl:163
 [2] ApplyArray(::Applied{LazyArrays.DefaultArrayApplyStyle,typeof(vec),Tuple{Array{Float64,2}}}) at /Users/sheehanolver/.julia/packages/LazyArrays/32TDf/src/lazyapplying.jl:164
 [3] ApplyArray(::Function, ::Array{Float64,2}) at /Users/sheehanolver/.julia/packages/LazyArrays/32TDf/src/lazyapplying.jl:168
 [4] copy(::ApplyArray{Float64,1,typeof(vec),Tuple{Array{Float64,2}}}) at /Users/sheehanolver/.julia/packages/LazyArrays/32TDf/src/lazyapplying.jl:187
 [5] top-level scope at REPL[14]:1
```

dlfivefifty:
> I'm not familiar with the code, please let me know if it should be changed elsewhere for instance 
> `copy(A::Applied{LazyArrayApplyStyle}) = ApplyArray(A)`

The code is largely designed to mimic the `Base.Broadcast` setup: a lazy `Applied` object is created, then materialized, which consists of calling  `initialize` and `copy`. `Applied` is too general to know `N` apriori, hence for each function `ndims` needs to overloaded. In short: I don't think this line should or can change as its not possible to know if its a `ApplyMatrix` or `ApplyVector` from the type signature of `Applied`.

tlienart:
added a simple test that would fail without the fix (btw, I've also tested locally that this `copy` effectively fixes the issue reported in https://github.com/JuliaData/Tables.jl/issues/122).

dlfivefifty:
Bump the patch number and I can tag right away

tlienart:
Done, thanks!

PR #90, opened 2019-12-22
Use $ to eagerly evaluate expressions in @~

tkf:
With this PR, you can use `@~ A .+ $Ref(x)` to eagerly evaluate the sub-expression `Ref(x)`.

### DIFF
--- a/src/lazymacro.jl
+++ b/src/lazymacro.jl
@@ -15,10 +15,17 @@ Broadcast.materialize(x::LazyCast) = x.value
 
 is_call(ex) = isexpr(ex, :call) && !is_dotcall(ex)
 
-is_dotcall(ex) =
-    (isexpr(ex, :.) && isexpr(ex.args[2], :tuple)) ||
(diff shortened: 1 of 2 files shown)


tkf:
This test is failing:

https://github.com/JuliaArrays/LazyArrays.jl/blob/7336bc5a95ce8118300f890479756331588a15ad/test/lazymultests.jl#L165

https://travis-ci.org/JuliaArrays/LazyArrays.jl/jobs/628247559#L302

But I don't understand why changing the macro definition would change this test. Also, I can't reproduce this locally with `Pkg.test(coverage=true)`... Any idea how to fix this?

dlfivefifty:
Just change it to `≈`: there's no reason to assume it should be `==`: `Array` multiplication uses BLAS while the other doesn't.

I have no idea why an unrelated change would trigger this, and only on two builds though.

tkf:
Thanks, using `≈` helps.

dlfivefifty:
Let me know when you want this merged

tkf:
It's good to go for me.

PR #94, opened 2020-01-14
Kron improvements

### LINKED ISSUE #65: Performance improvement of Kron times Vector/Matrix

emerali:
Resolves #65. 

Added materialize method for fast `Kron * AbstractVector` and `Kron * AbstractMatrix`, following the method detailed in this paper: http://yoksis.bilkent.edu.tr/pdf/files/11666.pdf (algorithms 1 & 3).
Also, a specialized `Kron * AbstractVector` for the case where the kronecker product only has two factors (the so-called "Vec Trick").

Minor additions: `logdet`, `logabsdet`, `diag`

Todo: 
- [x] tests
- [x] implement compatibility of Inv, Pinv, Adjoint and Transpose with Kron 
- [x] skip "modified shuffle" algorithm's initial step if the Kronecker factors can't be indexed/are lazy themselves ~~(need to think about how to implement this)~~

Possible todos:
- [x] another method for materialize in the case where both multiplication factors are Kron types
    - complication: need to check that factors are "aligned" before doing the factor-factor matmuls
- KronPow (matmul for this type [in the square case] ends up being a simple generalization of the Fast Walsh-Hadamard Transform and is therefore quite fast) and KronSum types?
    - admittedly, these may be outside of intended scope of project


src/lazyoperations.jl
```diff
@@ -59,6 +100,181 @@ function tr(K::Kron{<:Any, 2})
...
+        d = similar(K.args[1], minimum(size(K)))
+        for i in 1:length(d)
+            d[i] = K[i,i]
+        end
+        return d
+    end
+end
+
+
+function LazyArrays.materialize(M::Applied{
```
dlfivefifty: No need for `LazyArrays.` here and below

emerali: oops! this was a hold-over from when i was playing around with this in jupyter :joy:

emerali:
Hi there, I ended up getting busy with school work, so wasn't really able to keep up with this.

The PR as it is now *should* be OK to merge (modulo merge conflicts), but I have a couple questions/comments first

1. When multiplying 2 Kron types, the way I've implemented it now, materialize constructs the full matrix and returns that. This is because, as I understand it, the convention is for materialize to return a non-lazy array. Is my understanding correct? I think it'd make more sense in this case for Kron * Kron to give another Kron, so you can keep using the other optimized Kron methods with the result. I'm not sure how to do this without breaking the convention for materialize though.

2. For things like KronPow and KronSum, we'd need to make more types, and I'm worried this'll be straying out of the scope of LazyArrays. It might make more sense to spin Kronecker functionality out into a separate package entirely (we could still merge the current PR into LazyArrays for the time being, though). I came across a dedicated Kronecker package a while back (https://github.com/MichielStock/Kronecker.jl) that seems to do this, but is fairly rudimentary (eg. defines its structs recursively, making things like the shuffle algorithm difficult to implement). I did play around with it a little, and it *seems* to already be fairly compatible with LazyArrays since it's just a subtype of `AbstractMatrix`. Still, I'd like to know your thoughts on this.

dlfivefifty:
> the convention is for materialize to return a non-lazy array

No this isn't the convention. Return a lazy array if that's appropriate.

> For things like KronPow and KronSum, we'd need to make more types, and I'm worried this'll be straying out of the scope of LazyArrays

I'm not sure I understand what these types are meant to do.

emerali:
> Return a lazy array if that's appropriate.

this change has been made for compatible Kron by Kron matmuls

---

KronPow would just be an n-fold Kronecker product of a single matrix with itself

KronSum is definition 8 from: https://en.wikipedia.org/wiki/Kronecker_product#Relations_to_other_matrix_operations

They aren't too important though, so I won't add them in this PR (also I just don't think I'll have the time to implement them anytime soon).

dlfivefifty:
Do you know why the tests are passing on Julia v1.4?

src/lazyoperations.jl
```diff
@@ -41,24 +55,272 @@ function det(K::Kron{<:Any, 2})
...
+    # mixed-product property after explicitly computing the Kronecker product of
+    # the two factors of B (or maybe just deferring to the shuffle algorithm)
+    if (length(A.args) == length(B.args)) && all(_compatible_sizes, zip(A.args, B.args))
+        factors = [(A_i * B_i) for (A_i, B_i) in zip(A.args, B.args)]
+        return Kron(factors...)
+    else
+        algo_type = shuffle_algorithm_type(MT)
+        return shuffle_algorithm(algo_type, A, B, eltype(M))
+    end
+end
```
dlfivefifty, line 135: The following would be better to embrace the `MulAdd` type from ArrayLayouts.jl., to support in-place operations. For the case `(length(A.args) == length(B.args)) && all(_compatible_sizes, zip(A.args, B.args))` this seems fairly straightforward, something like:
```julia
similar(M::MulAdd{ApplyLayout{typeof(kron)}, ApplyLayout{typeof(kron)}, ::Type{T}) where T = Kron(similar.(MulAdd.(arguments(M.A), arguments(M.B)), T)...)

function materialize!(M:MulAdd{ApplyLayout{typeof(kron)}, ApplyLayout{typeof(kron)},ApplyLayout{typeof(kron)}})
   materialize!.(MulAdd.(M.α, arguments(M.A), arguments(M.B), M.β, arguments(M.C))
   M.C
end
```

Though I don't know how hard the other case would be... so happy to merge as is.

Technically you should be making a new `ApplyStyle` and not overload with `MulAddStyle`.

emerali: I don't understand the different `ApplyStyle`s or the types in ArrayLayouts.jl very well, so I'm not sure I'll be able to figure this out without help.

emerali:
I'm not sure if the tests are even actually failing in Julia 1.4

It looks like they're just timing out on Travis for some reason, but the timeout seems to be happening before the Kron tests are even reached as far as i can tell.

emerali:
hmm seems like there's a stack overflow happening, and i'm able to reproduce it locally too

i'll look into this

dlfivefifty:
If you get stuck I can have a look: parts of the design of this package are unfortunately brittle (in that if you don't overload the right things you get stack overflows), which I'm slowly correcting.

emerali:
I just gave it another attempt but i'm still unable to figure out what's causing the stack-overflow.

In addition to this, it appears that the materialize functions i defined are no longer being called. Not really sure how to fix them.

dlfivefifty:
OK I'll have a look

dlfivefifty:
OK I updated to the ArrayLayouts v0.4 system, so instead of `materialize(::Applied)` you overload `copy(::Mul)`.

emerali:
Tests seem to be passing locally now and I've made sure the correct methods are being called.

need to wait to see what Travis thinks

src/lazyoperations.jl
```diff
@@ -115,7 +115,12 @@ _compatible_sizes((A, B)) = (size(A, 2) == size(B, 1))
 
 # Implements the mixed-product property for the Kronecker product and matrix
 # multiplication
-function copy(M::Mul{ApplyLayout{typeof(kron)},ApplyLayout{typeof(kron)},<:AbstractMatrix,<:AbstractMatrix})
+function copy(M::Mul{
```
dlfivefifty, line 118: Why did you change the type signature? There is no reason to restrict to `Kron` as the `ApplyLayout` takes care of that/

dlfivefifty (changes requested):

src/lazyoperations.jl
emerali: i need to dispatch on the fact that both arguments here are `Kron` types or else it'll just go straight to the shuffle algo

dlfivefifty: No the first two `ApplyLayout{typeof(kron)}` tell you both are lazy kronecker products, you don't need to worry about the types of matrices.

emerali: oh you're right 

I think I got carried away when I fixed the vec-trick dispatch and just applied the same strategy here

emerali:
i'll write a few more tests to get the coverage back up

emerali:
ready to merge when you are

Project.toml
```diff
@@ -5,9 +5,11 @@ version = "0.17.5"
 [deps]
 ArrayLayouts = "4c555306-a7a7-4459-81d9-ec55ddd5c99a"
 FillArrays = "1a297f60-69ca-5386-bcde-b61e274b549b"
+Future = "9fa8497b-333b-5362-9e8d-4d0656e87820"
```
dlfivefifty, line 8: Why is this added?

src/lazyoperations.jl
```diff
@@ -41,24 +55,269 @@
...
+# multiplication
+function copy(M::Mul{ApplyLayout{typeof(kron)}, ApplyLayout{typeof(kron)}})
+    A, B = M.A, M.B
+    MT = promote_type(map(typeof, arguments(A))..., map(typeof,arguments(B))...)
+    # Keeping it simple for now, but could potentially make this "alignment"-check
+    # more precise. For example, if one factor of A has 6 columns and it's aligned
+    # with two factors of B with 2 and 3 rows, then we can still use the
+    # mixed-product property after explicitly computing the Kronecker product of
+    # the two factors of B (or maybe just deferring to the shuffle algorithm)
+    if (length(A.args) == length(B.args)) && all(_compatible_sizes, zip(A.args, B.args))
```
dlfivefifty, line 126: Use ` arguments(A) ` instead of ` A.args ` as ` A ` may potentially be a different type. Same with ` B.args `

src/lazyoperations.jl
```diff
@@ -41,24 +55,269 @@
...
+        return shuffle_algorithm(algo_type, A, B, eltype(M))
+    end
+end
+
+
+# Implements Roth's lemma aka the "vec-trick" for multiplying a 2-factor Kron
+# matrix to a vector
+function copy(M::Mul{
+    LazyArrays.ApplyLayout{typeof(kron)},
+    <:ArrayLayouts.MemoryLayout,
```
dlfivefifty, line 140: Why this restriction?

dlfivefifty (changes requested):

Project.toml
emerali: not really sure, i'll remove it

src/lazyoperations.jl
emerali: should i make this change everywhere, or just this function?

src/lazyoperations.jl
emerali: changed ArrayLayouts.MemoryLayout back to Any

dlfivefifty (approved):

PR #121, opened 2020-07-17
Redesign Mul

dlfivefifty:
This is a broad PR that redesigns `Mul`. Actually, `Mul` is now in ArrayLayouts.jl, and is specifically a 2-argument multiplication. (This then determines whether to use `MulAdd`, `Lmul` or `Rmul` depending on the `MemoryLayout`.)

The benefit of this is we get to completely remove the `@lazymul` macro as the 2-argument multiplication now follows ArrayLayouts.jl design. This also means much fewer styles.


dlfivefifty:
@tkf if you have time to comment/discuss this PR it would be great. 

The main goal of this is to remove the `@lazymul` macro, and instead for 2-argument multiplication use `MulStyle` to lower to `ArrayLayouts.Mul(A,B)` which uses `MemoryLayout` to determine how it is materialised. This should simplify things, and removes a lot of notions such as `LmulStyle` and `RmulStyle`. 

`combine_mul_styles` was also redesigned not to use `MemoryLayout`, with the idea to keep memory layouts as disconnected from `ApplyStyle` as possible (for more sane code modularity). To be honest I ended up hacking something together, partly so that QuasiArrays.jl can successfully take over if an `AbstractQuasiArray` is in the mix.

dlfivefifty:
This is in some sense a reversion to an earlier design but with some extra layers in the mix. That is, there is sort of a tree of type lowering: `Applied(*, A, B)` -> `Mul(A,B)` -> `MulAdd(1.0, A, B, 0.0, C)`, `Lmul(A, B)` or `Rmul(A, B)` depending on the `MemoryLayout` of `A` and `B`. This design helps mitigate some of the annoying type ambiguities of one layer, (hopefully) in a sane manner.

Issue #132, opened 2020-08-30
Indexing performance for heterogeneous concatenated arrays


jipolanco:
Indexing heterogeneous concatenated arrays is slow due to type instability issues. By heterogeneous I mean something like:

```julia
A = Vcat(
    rand(1:10, 12),
    @SVector(zeros(Int, 4)),
    4:42,
)
```

where each concatenated subarray may have different type.

Indexing the first subarray is relatively fast, while indexing subsequent arrays becomes slower and allocates memory. On Julia 1.5.1:

```julia
N = length(A)
@btime $A[1];   # 7.356 ns (0 allocations: 0 bytes)
@btime $A[14];  # 43.875 ns (2 allocations: 112 bytes)
@btime $A[$N];  # 80.788 ns (4 allocations: 208 bytes)
```

This seems to be caused by type instabilities in `vcat_getindex`, as can be checked with `@code_warntype LazyArrays.vcat_getindex(A, N)`. More precisely, the problem is in the loop:
```julia
    for A in f.args
        n = length(A)
        κ ≤ n && return convert(T,A[κ])::T
        κ -= n
    end
```
where `A` can be any of the concatenated subarrays.

The same issue occurs with `setindex!`, and also with two-dimensional concatenated arrays (`Vcat{2}` and `Hcat`).

I've been playing with a possible fix using recursive iteration over the `f.args` tuple, and I'll submit a PR soon.

jipolanco:
Closed by #133

PR #133, opened 2020-08-30
Recursive indexing of concatenated arrays

jipolanco:
This is an attempt to fix the performance issues discussed in #132. These are type instability issues when indexing heterogeneous concatenated arrays with `getindex` and `setindex!`.

In this PR, `getindex` and `setindex!` are reimplemented for the `Vcat{1}`, `Vcat{2}` and `Hcat` lazy concatenation types.

Here is a small benchmark for comparison, on Julia 1.5.1:

```julia
using BenchmarkTools
using LazyArrays
using StaticArrays

A = Vcat(
    rand(1:10, 12),
    @SVector(zeros(Int, 4)),
    4:42,
)

N = length(A)
```

On LazyArrays master:
```julia
@btime $A[1];   # 7.356 ns (0 allocations: 0 bytes)
@btime $A[14];  # 43.875 ns (2 allocations: 112 bytes)
@btime $A[$N];  # 80.788 ns (4 allocations: 208 bytes)
```

With this PR:
```julia
@btime $A[1];   # 4.329 ns (0 allocations: 0 bytes)
@btime $A[14];  # 4.668 ns (0 allocations: 0 bytes)
@btime $A[$N];  # 6.105 ns (0 allocations: 0 bytes)
```

### DIFF
--- a/src/lazyconcat.jl
+++ b/src/lazyconcat.jl
@@ -43,27 +43,31 @@ function ==(a::Vcat{T,1,II}, b::Vcat{T,1,II}) where {T,II}
     all(arguments(a) .== arguments(b))
 end
 
-@propagate_inbounds @inline function vcat_getindex(f, k::Integer)
+@propagate_inbounds @inline vcat_getindex(f, k::Integer) =
+    vcat_getindex_recursive(f, (k, ), f.args...)
+
+@propagate_inbounds @inline vcat_getindex(f, k::Integer, j::Integer) =
+    vcat_getindex_recursive(f, (k, j), f.args...)
+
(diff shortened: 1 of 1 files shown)


dlfivefifty:
Looks like just one line is missing in coverage, are you able to cover this? (Sometimes its just a bug in codecov and I can merge as is in that case.)

jipolanco:
Sorry, I had left a function that in the end I didn't use. Should be good now.

dlfivefifty:
If bump the patch version (to v0.17.7) I can merge and tag right away

dlfivefifty:
Actually I'll just do the tag now

PR #139, opened 2020-10-21
Workaround type inference issue on Julia master

### LINKED ISSUE #138: Type inference issues on Julia master (v1.6)
The following is not inferred on Julia master (but works correctly on 1.5):

```julia
x = rand(2, 5)
@code_warntype ApplyArray(vcat, x)

# Output:
Variables
  #self#::Type{ApplyArray}
  f::Core.Const(vcat)
  factors::Tuple{Matrix{Float64}}

Body::Vcat{Float64, _A, Tuple{Matrix{Float64}}} where _A
1 ─ %1 = Core.tuple(f)::Core.Const((vcat,))
│   %2 = Core._apply_iterate(Base.iterate, LazyArrays.applied, %1, factors)::Applied{LazyArrays.DefaultArrayApplyStyle, typeof(vcat), Tuple{Matrix{Float64}}}
│   %3 = LazyArrays.ApplyArray(%2)::Vcat{Float64, _A, Tuple{Matrix{Float64}}} where _A
└──      return %3
```

The critical step seems to be around here:
```julia
M = applied(vcat, x)
@inferred ApplyArray(M)                       # fails
@inferred ApplyArray{eltype(M)}(M)            # fails
@inferred ApplyArray{eltype(M), ndims(M)}(M)  # works
```

I'll make a PR with a workaround that seems to fix this issue.

jipolanco:
Fixes #138

### DIFF
--- a/src/lazyconcat.jl
+++ b/src/lazyconcat.jl
@@ -28,7 +28,7 @@ end
 
 @inline eltype(A::Applied{<:Any,typeof(vcat)}) = promote_type(map(eltype,A.args)...)
 @inline eltype(A::Applied{<:Any,typeof(vcat),Tuple{}}) = Any
-@inline ndims(A::Applied{<:Any,typeof(vcat),I}) where I = max(1, max(map(ndims,A.args)...))
+@inline ndims(A::Applied{<:Any,typeof(vcat),I}) where I = max(1, map(ndims,A.args)...)
 @inline ndims(A::Applied{<:Any,typeof(vcat),Tuple{}}) = 1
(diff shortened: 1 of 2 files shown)


dlfivefifty:
Thanks! Do you know why this regression happened? Maybe we should file an issue if it can be reproduced on a simpler example?

jipolanco:
You're right, it may be a regression in base Julia. I'll see if I can reproduce a minimal example and file an issue.

jipolanco:
Closing this because inference regression was fixed on Julia master.

PR #173, opened 2021-06-01
Support * for BroadcastArray(+, ....)

dlfivefifty:


dlfivefifty:
@tkf if you have any comments on this that would be greatly appreciated.

Note there's a bit of a disconnect between functionality of `ApplyArray(+, ...)` and `BroadcastArray(+, ...)`. This is partly due to the challenge in supporting `muladd!` in the latter, but more accurately because of what I need downstream.

PR #181, opened 2021-07-03
Lazy ndgrid

JeffFessler:
Here is an initial PR for adding lazy `ndgrid`.
Some questions:

Currently there is `ndgrid` and `ndgrid_lazy` method names.
If you prefer `ndgrid` for the lazy version of course that's fine.
Having the non-lazy version (with some name, somewhere) is helpful for testing.
I don't think we need the capitalized `Ndgrid` here because Base will *never* offer `ndgrid` :)

I'm unsure about bounds checking so please double check all that.

Also, if you have any ideas about how to accelerate it, please advise!
There are (commented out) timing tests in the test code that are about 4x slower which is tolerable, but any improvement would be welcome.  My guess is that `r.args[1][i[r.args[3]]]` may simply have that much overhead...

Do you want `meshgrid` too?  It's just one more line:
`meshgrid(y,x) = (ndgrid_lazy(x,y)[[2,1]]...,)`

If this gets merged we should advertise it:
https://discourse.julialang.org/t/meshgrid-function-in-julia/48679
https://stackoverflow.com/questions/44581049/utilizing-ndgrid-meshgrid-functionality-in-julia/44766347

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "LazyArrays"
 uuid = "5078a376-72f3-5289-bfd5-ec5146d43c02"
-version = "0.21.9"
+version = "0.22.0"
 
 [deps]
 ArrayLayouts = "4c555306-a7a7-4459-81d9-ec55ddd5c99a"
--- a/src/LazyArrays.jl
+++ b/src/LazyArrays.jl
@@ -76,6 +76,7 @@ include("lazyconcat.jl")
 include("lazysetoperations.jl")
 include("lazyoperations.jl")
 include("lazymacro.jl")
+include("ndgrid.jl")
 
 # support x^2
 Base.to_power_type(x::LazyArray) = x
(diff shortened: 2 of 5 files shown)


dlfivefifty:
What would you say to putting this in a separate package called LazyGrids.jl? I think that would make it more visible and we could add other lazy grids there 

https://github.com/JuliaApproximation/FastTransforms.jl/blob/feb2c18fc2b1ba1e2a7a60feb04289fc17d7b4e6/src/chebyshevtransform.jl#L393

JeffFessler:
That would be fine with me.  The logical place for it would seem to be under JuliaArrays, right?
What's the best way to proceed?

dlfivefifty:
I can make a repos at add you as contributor

dlfivefifty:
I've invited you to join 

https://github.com/JuliaArrays/LazyGrids.jl

JeffFessler:
Closing this PR because I have much cleaner version in progress at LazyGrids.  Thanks!

PR #205, opened 2022-03-05
Fix tests on Julia 1.8-beta1

rikhuijzer:
I've tested these changes on Julia 1.6, 1.7 and 1.8-beta1.

I've also changed

```
on:
  - push
  - pull_request
```
to
```
on:
  push:
    branches:
      - master
  pull_request:
```
to avoid triggering the jobs for pushes and pull requests at the same time. See, for example, https://github.com/JuliaData/CSV.jl/blob/2c02394b4ce11ee673ed1b6d0c811a98316bc2f9/.github/workflows/ci.yml#L3-L5

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,8 @@
 name: CI
 on:
-  - push
+  - push:
+      branches:
+        - main
   - pull_request
 jobs:
   test:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,7 @@ name: CI
 on:
   push:
     branches:
-      - main
+      - master
   pull_request:
 jobs:
   test:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,8 @@ jobs:
       matrix:
         version:
           - '1.6'
-          - '^1.7.0-0'
+          - '1' # Latest stable release
+          - '^1.8.0-0'
         os:
           - ubuntu-latest
           - macOS-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,9 +1,9 @@
 name: CI
 on:
-  - push:
-      branches:
-        - master
-  - pull_request:
+  push:
+    branches:
+      - main
+  pull_request:
 jobs:
   test:
     name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ github.event_name }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,7 @@ on:
   - push:
       branches:
         - master
-  - pull_request
+  - pull_request:
 jobs:
   test:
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -1,6 +1,8 @@
 name: CI
 on:
-  - push
+  - push:
+      branches:
+        - main
```
DilumAluthge, line 5:
```suggestion
        - master
```

@mbauman 

h/t @giordano

test/applytests.jl
```diff
@@ -58,7 +58,11 @@ import ArrayLayouts: StridedLayout
         @test copyto!(b, a) == a == b
         @test copyto!(c, a) == a == c
         @test copyto!(similar(a), a) == a
-        @test_throws ErrorException copyto!(c, Array(a))
+        if VERSION ≤ v"1.7.99"
```
giordano, line 61:
```suggestion
        if VERSION ≤ v"1.8.0-DEV"
```

DilumAluthge: This would also work:

```
if VERSION < v"1.8.0-DEV"
```

rikhuijzer: Great! I knew there was a better way. Thanks

.github/workflows/ci.yml
```diff
@@ -1,7 +1,9 @@
 name: CI
 on:
-  - push
-  - pull_request
+  push:
+    branches:
+      - main
```
giordano:
```suggestion
      - master
```
Again

rikhuijzer: Sorry. Fixed in dcb54c7

rikhuijzer:
Can someone approve the workflow?

dlfivefifty:
The fixes look like they are working around regressions so I don't think this should be merged as is. E.g. replacing `≡` with `==` in
```julia
@test apply(*, Eye(10), Ones(10)) ≡ Ones(10)
```
is an issue: we want it to be identically a `Ones`, not just have the same entries.

rikhuijzer:
> is an issue: we want it to be identically a `Ones`, not just have the same entries.

Do you have a suggestion on how to fix it? I think that the failures are due to the new representations for sparse matrices

dlfivefifty:
Just need to step through the debugger and see what's going wrong

test/paddedtests.jl
```diff
@@ -207,6 +207,7 @@ paddeddata(a::PaddedPadded) = a
             @test a .* v == v .* a
 
             @test n .+ n ≡ Vcat(2,Zeros{Int}(3))
+            @test n .+ v ≡ n .+ v
```
rikhuijzer, line 210: I've added this one because currently this doesn't even hold on Julia 1.8-beta1. Testing this may save a lot of time in the future when trying to figure out why `n .+ v !== v .+ n`.

rikhuijzer:
Tests are still broken. Patch for FillArrays is at https://github.com/JuliaArrays/FillArrays.jl/pull/172 and I'm working on a patch for FillArrays too. Those two patches will fix the tests here. 

At FillArrays,

```julia
julia> ArrayLayouts.mul(Eye{Int}(10), Eye(10)) ≡ Eye(10)
false
```
which should be `true`.

rikhuijzer:
Awesome! Thanks for fixing this and MatrixFactorizations, @dlfivefifty

PR #243, opened 2023-03-21
fix typos

spaette:
src/cache.jl
src/lazybroadcasting.jl
src/lazyconcat.jl
src/linalg/add.jl
test/broadcasttests.jl
test/concattests.jl
test/multests.jl


spaette:
@dlfivefifty 

Please clarify whether the JuliaArray Organization would like reviews of specific other repositories.

PR #260, opened 2023-07-14
support Julia 1.9

putianyi889:


dlfivefifty:
These changes aren't needed since testing on '1' is the same as v1.9 and writing `julia = "1.6"` means `1.6 ≤ VERSION < 2` which includes 1.9

Issue #255, opened 2023-07-07
REPL dies to repeat ApplyArray multiplications


TSGut:
@dlfivefifty Continuing our discussion in Github:

This is related to fixing the issue in https://github.com/JuliaApproximation/SemiclassicalOrthogonalPolynomials.jl/issues/78
I will use ClassicalOPs.jl for my example instead since it keeps it simpler. 

Basically, this kills the REPL:
```julia
using ClassicalOrthogonalPolynomials, LinearAlgebra, LazyArrays
P = Legendre()
X = UpperTriangular(I+jacobimatrix(P))

Aj = X
for j = 1:4
    Aj = ApplyArray(*,Aj,Aj)
end
Aj
```

but it looks to me to be specifically relating to the type display since I can evaluate the entries just fine, e.g.
```julia
using ClassicalOrthogonalPolynomials, LinearAlgebra, LazyArrays
P = Legendre()
X = UpperTriangular(I+jacobimatrix(P))

Aj = X
for j = 1:4
    Aj = ApplyArray(*,Aj,Aj)
end
Aj[1:50,1:50]
```
works just fine.

Since the entries work, one way around this would be for me to just create some new struct `HiddenTypeArray{T}` or whatever and just have it defer all operations to its data. Is there a more elegant solution?

TSGut:
Changed the example to be a slightly less degenerate case.

dlfivefifty:
Hit control C and see where it hangs

putianyi889:
After #261, REPL outputs
```julia
[...]

PR #261, opened 2023-07-15
Accelerate `_mul_ApplyStyle`

putianyi889:
Before:
```julia
julia> @time ApplyArray(*, [ones(10,10) for k in 1:25]...);
105.543375 seconds (92.57 M allocations: 3.603 GiB, 1.06% gc time, 0.35% compilation time)
```

After:
```julia
julia> @time ApplyArray(*, [ones(10,10) for k in 1:25]...);
  0.163698 seconds (52.06 k allocations: 3.539 MiB, 98.30% compilation time)

julia> @time ApplyArray(*, [ones(10,10) for k in 1:200]...);
  1.248202 seconds (660.75 k allocations: 48.919 MiB, 8.38% gc time, 97.46% compilation time)
```

Also happens to solve #255 

Note that the compilation time will always kick in.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 deps/deps.jl
 .DS_Store
 Manifest.toml
+statprof/*
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "LazyArrays"
 uuid = "5078a376-72f3-5289-bfd5-ec5146d43c02"
-version = "1.4.0"
+version = "1.4.1"
 
 [deps]
 ArrayLayouts = "4c555306-a7a7-4459-81d9-ec55ddd5c99a"
--- a/src/linalg/mul.jl
+++ b/src/linalg/mul.jl
@@ -92,7 +92,9 @@ combine_mul_styles(a, b, c...) = combine_mul_styles(combine_mul_styles(a, b), c.
 # We need to combine all branches to determine whether it can be  simplified
 ApplyStyle(::typeof(*), a) = DefaultApplyStyle()
 ApplyStyle(::typeof(*), a::AbstractArray) = DefaultArrayApplyStyle()
-_mul_ApplyStyle(a, b...) = combine_mul_styles(ApplyStyle(*, a, Base.front(b)...),  ApplyStyle(*, b...))
+_mul_ApplyStyle(a, b...) = combine_mul_styles(_mul_ApplyStyle(a, Base.front(b)...),  _mul_ApplyStyle(b...))
+_mul_ApplyStyle(a, b) = MulStyle()
+_mul_ApplyStyle(a) = MulStyle()
 ApplyStyle(::typeof(*), a, b...) = _mul_ApplyStyle(a, b...)
 if !(AbstractQ  <: AbstractMatrix) # VERSION >= v"1.10-"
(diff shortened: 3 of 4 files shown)


dlfivefifty:
Amazing, thanks for this!!

dlfivefifty:
Oh wait it’s doing things at run time. I think we should make it a generated function to do this at compile time

putianyi889:
I had thought about that for a while but couldn't come up with an idea.

putianyi889:
alright, just adding `@generated` does the job.

One-line recursion:
```julia
julia> using LazyArrays

julia> A=[ones(10,10) for k in 1:500];

julia> @time ApplyArray(*, A...);
 28.553755 seconds (2.82 M allocations: 549.886 MiB, 0.82% gc time, 99.56% compilation time)

julia> @time ApplyArray(*, A...);
  0.087592 seconds (10.59 k allocations: 6.566 MiB)

julia> B=A[1:200];

julia> @time ApplyArray(*, B...);
  4.429628 seconds (125.22 k allocations: 27.461 MiB, 0.63% gc time, 99.54% compilation time)

julia> @time ApplyArray(*, B...);
  0.016024 seconds (4.29 k allocations: 1.219 MiB)
```

$O(n^2)$ algorithm:
```julia
julia> using LazyArrays
[ Info: Precompiling LazyArrays [5078a376-72f3-5289-bfd5-ec5146d43c02]

julia> A=[ones(10,10) for k in 1:500];

julia> @time ApplyArray(*, A...);
  4.793488 seconds (2.13 M allocations: 172.483 MiB, 1.48% gc time, 97.39% compilation time)

julia> @time ApplyArray(*, A...);
  0.085763 seconds (10.59 k allocations: 6.566 MiB)

julia> B=A[1:200];

julia> @time ApplyArray(*, B...);
  0.645032 seconds (53.70 k allocations: 3.771 MiB, 96.96% compilation time)

julia> @time ApplyArray(*, B...);
  0.016678 seconds (4.29 k allocations: 1.219 MiB)
```

dlfivefifty:
I don't understand how this is working: your implementation of `_mul_ApplyStyle` doesn't actually return anything?

putianyi889:
the last assignment which is list[1] is returned

putianyi889:
it was a mistake... I have no idea why nothing went wrong

putianyi889:
can you approve the workflows? @dlfivefifty

dlfivefifty:
Can you add tests with `@inferred` and `@allocated` to make sure we are actually doing things at compile-time?

You can confirm this yourself as well using `@code_warntype ApplyStyle(*, a, b)`, which will just return the applystyle with no actual code

putianyi889:
I don't understand how they work. If they take run-time, will `@inferred` throw an error? what about `@allocated`?

dlfivefifty:
Since the only thing it actually is computing is a return type, by definition of the code is inferred then its doing everything at compile time. 


```julia
julia> @inferred(ApplyStyle(*, A, A))
LazyArrays.DefaultArrayApplyStyle()
```

The allocations is just a sanity check but they are too unreliable so forget that. Just add some `@inferred` to some `ApplyStyle` calls and that should be enough.

dlfivefifty:
ignore codecov since its just because generated functions are trackable

putianyi889:
Is that enough?

dlfivefifty:
I'll merge this once tests pass.

If you want a PR tagged remember to bump the version number

putianyi889:
version bumped

PR #265, opened 2023-07-18
cooperate with ArrayLayouts #153

putianyi889:
https://github.com/JuliaLinearAlgebra/ArrayLayouts.jl/pull/153

fix ambiguities in tests


putianyi889:
can you merge? @dlfivefifty

dlfivefifty:
No because I don't agree with the proposed change to ArrayLayouts.jl so this shouldn't be needed

PR #268, opened 2023-08-20
Improve inlining, faster Lazy mul getindex

dlfivefifty:


dlfivefifty:
@putianyi889 I think this supersedes https://github.com/JuliaLinearAlgebra/ArrayLayouts.jl/pull/153

dlfivefifty:
These changes are not technically breaking but trigger bugs in downstream packages that are only fixed on versions require Julia v1.9. I think its therefore reasonable to drop support for older versions of Julia.

Issue #314, opened 2024-06-04
Downstream testing for InfiniteLinearAlgebra.jl


DanielVandH:
https://github.com/JuliaArrays/LazyArrays.jl/blob/052a0f6cd85911951b3ec6a6b6030420926d70f6/.github/workflows/downstream.yml#L34

Does this still need to be commented out?

dlfivefifty:
Oh forgot this was commented out. Yes we can turn the tests back on

PR #315, opened 2024-06-04
Update downstream.yml to include InfiniteLinearAlgebra.jl again

DanielVandH:
Fixes https://github.com/JuliaArrays/LazyArrays.jl/issues/314


dlfivefifty:
Do you understand why the tests fail? It looks like it’s using the latest version of InfiniteLinearAlgebra.jl which I thought fixed this..

dlfivefifty:
oh wait... the tests on inflinalg are actually broken...

PR #319, opened 2024-06-18
Add missing `const` declarations

### LINKED ISSUE #318: missing `const`s in extension source code
In the extensions, there are some examples of variables defined without a `const`, e.g. 

https://github.com/JuliaArrays/LazyArrays.jl/blob/fc9e0dca2c35e21f429ecca924bacb6b30eeb9ce/ext/LazyArraysBlockBandedMatricesExt.jl#L25-L27

This is probably causing some slight performance penalties (as far as my understanding goes? the third line in the above excerpt definitely should), since for example

```julia
module Ex1
export exmethod1
ind(x) = x
_ind = ind
exmethod1(x) = _ind(x) # bad 
end
module Ex2
export exmethod2
ind(x) = x
const _ind = ind
exmethod2(x) = _ind(x) # good
end

using .Ex1, .Ex2
```
```julia-repl
julia> @code_warntype exmethod1(1)
MethodInstance for Main.Ex1.exmethod1(::Int64)
  from exmethod1(x) @ Main.Ex1 Untitled-1:5
Arguments
  #self#::Core.Const(Main.Ex1.exmethod1)
  x::Int64
Body::Any
1 ─ %1 = Main.Ex1._ind(x)::Any
└──      return %1


julia> @code_warntype exmethod2(1)
MethodInstance for Main.Ex2.exmethod2(::Int64)
  from exmethod2(x) @ Main.Ex2 Untitled-1:11
Arguments
  #self#::Core.Const(Main.Ex2.exmethod2)
  x::Int64
Body::Int64
1 ─ %1 = Main.Ex2._ind(x)::Int64
└──      return %1
```

I can try and get around to fixing these eventually

DanielVandH:
Closes #318. Pretty sure I found all of them

### DIFF
--- a/ext/LazyArraysBandedMatricesExt.jl
+++ b/ext/LazyArraysBandedMatricesExt.jl
@@ -535,11 +535,11 @@ bandeddata(R::ApplyMatrix{<:Any,typeof(rot180)}) = @view(bandeddata(arguments(R)
 # overload copy as overloading `mulreduce` requires `copyto!` overloads
 # Should probably be redesigned in a trait-based way, but hard to see how to do this
 
-BandedLazyLayouts = Union{AbstractLazyBandedLayout, BandedColumns{LazyLayout}, BandedRows{LazyLayout},
-TriangularLayout{UPLO,UNIT,BandedRows{LazyLayout}} where {UPLO,UNIT},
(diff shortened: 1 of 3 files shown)


DanielVandH:
Not really sure what the issues are with those errors.. not sure how they could be related to these changes. I don't get them locally but I probably need to try with the exact test manifest to see what's going on. Julia is being a pain for me currently so I'll just have to come back to this.

PR #328, opened 2024-07-03
Add == for Zeros-backed CachedArrays

### LINKED ISSUE #327: cache(Zeros(n, n)) == cache(Zeros(n, n)) uses matrix ==
```julia
julia> using LazyArrays, FillArrays

julia> @which cache(Zeros(5)) == cache(Zeros(5)) # good
==(A::LazyArrays.CachedArray{<:Any, 1, <:Any, <:Zeros}, B::LazyArrays.CachedArray{<:Any, 1, <:Any, <:Zeros})
     @ LazyArrays C:\Users\djv23\.julia\packages\LazyArrays\44w6E\src\padded.jl:121

julia> @which cache(Zeros(5, 5)) == cache(Zeros(5, 5)) # fallback
==(A::AbstractArray, B::AbstractArray)
     @ Base abstractarray.jl:2937
```

Would be good to have them both be a specialised definition to make it easier to do `==` being `Zeros`-backed `cache`s. Will make a pr

DanielVandH:
Fixes #327

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "LazyArrays"
 uuid = "5078a376-72f3-5289-bfd5-ec5146d43c02"
-version = "2.1.2"
+version = "2.1.3"
 
 [deps]
 ArrayLayouts = "4c555306-a7a7-4459-81d9-ec55ddd5c99a"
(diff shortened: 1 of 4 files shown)


dlfivefifty:
To fix this properly we want to add an `equals_layout` to ArrayLayouts.jl a la 

https://github.com/JuliaApproximation/QuasiArrays.jl/blob/d9ce29e4cb188ec5b246aaae6242589a11d873c5/src/abstractquasiarray.jl#L506

Then we can compare any `AbstractPaddedLayout`. But we can merge this as is for now

PR #337, opened 2024-07-10
Change signature of `materialize!` to fix regression on Windows

DanielVandH:
As discussed here https://github.com/JuliaApproximation/SemiclassicalOrthogonalPolynomials.jl/pull/111, this fixes the weird inlining bug/regression on Windows by changing the signature of `materialize!`. Don't know if this is the best solution ideally, but it seems to work. Tests pass locally but let's see what the downstream CI says.

Don't really know the best way to a test of this regression to avoid this happening again.. I guess just checking that the Windows CI times are comparable to the other OSs?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "LazyArrays"
 uuid = "5078a376-72f3-5289-bfd5-ec5146d43c02"
-version = "2.1.6"
+version = "2.1.7"
 
 [deps]
 ArrayLayouts = "4c555306-a7a7-4459-81d9-ec55ddd5c99a"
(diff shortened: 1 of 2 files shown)


DanielVandH:
Seems to have worked, all the CI timings are independent of OS

DanielVandH:
> I don’t think the downstream tests will check on windows but hopefully it fixes the issue

Would it be worth putting 

```
        os:
          - ubuntu-latest
          - windows-latest
          - macos-latest
```

into the downstream.yml to test? That'd make a ton of workflows I guess is a problem

dlfivefifty:
No it’s not worth it. I’ll just merge and tag

PR #334, opened 2024-07-04
Add `(row/col)support` for `inv(Triangular)`

DanielVandH:

PR #338, opened 2024-07-11
Revert "Add (row/col)support for inv(Triangular)"

DanielVandH:
Reverts https://github.com/JuliaArrays/LazyArrays.jl/pull/334. See https://github.com/JuliaApproximation/SemiclassicalOrthogonalPolynomials.jl/pull/111.

Hopefully this method can be introduced once again eventually, but for now can we remove it?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "LazyArrays"
 uuid = "5078a376-72f3-5289-bfd5-ec5146d43c02"
-version = "2.1.7"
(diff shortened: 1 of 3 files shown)


dlfivefifty:
Oh cause those were just wrong: the inverse of a banded triangular matrix is triangular but not banded

DanielVandH:
Oh. Duh. I definitely should've noticed that...

PR #339, opened 2024-07-11
Add SemiclassicalOrthogonalPolynomials.jl to downstream tests

DanielVandH:
Will help prevent future issues with the $b=-1$ stuff especially @dlfivefifty (once that's merged anyway).


dlfivefifty:
Note I'm not sure its the best idea to add every downstream package but since this appears to have triggered an actual issue I've merged

PR #345, opened 2024-09-05
Avoid depending on `LazyArraysBandedMatricesExt` from within `LazyArraysBlockBandedMatricesExt`

topolarity:
LazyArrays reasonably assumes that an extension w/ triggers `{B,C}` to can depend on an extension w/ triggers `{A,C}` - as long as B depends on A.

Unfortunately as detailed in https://github.com/JuliaLang/julia/issues/55557, Julia's extension loading does not actually guarantee that. Extension loading is purely trigger-based, so if `C` is the loaded after `{A,B}` then both of these extensions will trigger "simultaneously" and they will be loaded in an indeterminate order. That can mean that the first extension loads after the second, even though it expected to depend on it.

That behavior is arguably a bug and is likely to be improved in a future version of Julia, but for now this works around the problem by removing the inter-extension dependency.


dlfivefifty:
Thanks for this!

PR #347, opened 2024-11-12
Remove backwards compatibility package extension logic

mtfishman:
Some weakdeps were listed as deps so that package extensions could be loaded in older Julia versions, which doesn't seem necessary now since LazyArrays.jl requires Julia 1.10. See also https://github.com/JuliaArrays/BlockArrays.jl/pull/431.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,17 +1,13 @@
 name = "LazyArrays"
 uuid = "5078a376-72f3-5289-bfd5-ec5146d43c02"
-version = "2.2.1"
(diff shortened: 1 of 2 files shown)


mtfishman:
As a more obvious downside of listing them as both dependencies and weak dependencies, doesn't that cause those packages (`BandedMatrices`, `BlockArrays`, `BlockBandedMatrices`, `StaticArrays`) to be downloaded and installed for every user, even if they don't need it?

PR #365, opened 2025-02-11
Move SparseArrays dependency to extension

j-fu:
This should help to resolve https://github.com/SciML/LinearSolve.jl/issues/573

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -7,18 +7,19 @@ ArrayLayouts = "4c555306-a7a7-4459-81d9-ec55ddd5c99a"
 FillArrays = "1a297f60-69ca-5386-bcde-b61e274b549b"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 MacroTools = "1914dd2f-81c6-5fcd-8719-6d5c9610ff09"
-SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 
 [weakdeps]
 BandedMatrices = "aae01518-5342-5314-be14-df237901396f"
 BlockArrays = "8e7c35d0-a365-5155-bbbb-fb81a777f24e"
 BlockBandedMatrices = "ffab5731-97b5-5995-9138-79e8c1846df0"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
+SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 
 [extensions]
 LazyArraysBandedMatricesExt = "BandedMatrices"
 LazyArraysBlockArraysExt = "BlockArrays"
 LazyArraysBlockBandedMatricesExt = "BlockBandedMatrices"
+LazyArraysSparseArraysExt = "SparseArrays"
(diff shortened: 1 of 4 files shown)


dlfivefifty:
"my_" is a very bad naming convention. I have no idea what these are for!

j-fu:
Well the idea is to have some issparse and nnz methods which defaults to false in the case SparseArrays is not loaded, and is extended with the ones from SparseArrays in the extensions. Do you agree with the general approach ?

Instead of "my_" we could use "_", "tentative_" "local_"  "provisional_" "preliminary_" ...

ChrisRackauckas:
@j-fu I think `local` makes sense. Can you go for that?

j-fu:
Review from Copilot/GPT5:

> 
> PR review: Move SparseArrays dependency to extension (#365)
> 
> Summary
> - Converts SparseArrays from a hard dependency to an extension.
> - Adds ext/LazyArraysSparseArraysExt.jl with methods that bridge to SparseArrays when available.
> - Introduces internal fallbacks local_issparse and local_nnz and switches internal usage from issparse/nnz to these local functions.
> - Updates Project.toml: moves SparseArrays to [weakdeps], adds LazyArraysSparseArraysExt in [extensions], adds compat entry and includes SparseArrays in [targets] test.
> - Removes SparseArrays from the main using list in src/LazyArrays.jl.
>
> Correctness and design
> - The fallback definitions
>  - local_issparse(A) = false
>  - local_nnz(A) = prod(size(A))
>  preserve the previous logic in lazyoperations: nnz(X) is only used when X is sparse; otherwise prod(size(X)) is used. With the extension loaded, local_issparse(::AbstractArray) delegates to issparse and local_nnz(::AbstractSparseArray) to nnz, restoring the prior behavior without requiring SparseArrays globally.
> - Dispatch is sensible: a broad Any fallback in-package and narrower AbstractArray/AbstractSparseArray methods in the extension. This avoids method piracy and keeps SparseArrays-specific code isolated.
> - The changes in shuffle_algorithm correctly swap any(issparse, ...) and nnz(...) uses with local_issparse/local_nnz, maintaining semantics.
> - Given LazyArrays already uses extensions for other packages, this is consistent with the package’s existing approach.

> Nits and small suggestions
> - src/LazyArrays.jl: consider removing the commented fragment in the using line to keep it clean:
>  - Current: using Base.Broadcast, LinearAlgebra, FillArrays, ArrayLayouts #, SparseArrays
>  - Suggested: using Base.Broadcast, LinearAlgebra, FillArrays, ArrayLayouts
> - In lazyoperations.jl, local_nnz(A) = length(A) would be equivalent and a touch clearer than prod(size(A)) for AbstractArray inputs; either is fine.
> - ext/LazyArraysSparseArraysExt.jl: a brief module comment noting the intent (bridge to SparseArrays for local_* functions) would mirror the helpful in-file comment in lazyoperations.jl.

> Overall
> - The implementation is minimal, well-scoped, and preserves behavior while removing the unconditional SparseArrays load. The approach looks solid and aligns with Julia’s extension mechanism. LGTM with the minor cleanups noted above.

dlfivefifty:
If you bump the minor version (1.x) I can merge and tag

PR #366, opened 2025-02-14
Explicitly import `ArrayLayouts.MemoryLayout`

nsajko:
Prevent name ambiguity, adapt to changes in nightly Julia, preventing a warning during precompilation.

See:
* https://github.com/JuliaLang/julia/issues/25744
* https://github.com/JuliaLang/julia/issues/57290
* https://github.com/JuliaLang/julia/pull/57311

### DIFF
--- a/src/LazyArrays.jl
+++ b/src/LazyArrays.jl
@@ -5,7 +5,6 @@ module LazyArrays
 
 using Base.Broadcast, LinearAlgebra, FillArrays, ArrayLayouts, SparseArrays
 import LinearAlgebra.BLAS
-import ArrayLayouts.MemoryLayout
 
 import Base: *, +, -, /, <, ==, >, \, ≤, ≥, (:), @_gc_preserve_begin, @_gc_preserve_end, @propagate_inbounds,
              AbstractArray, AbstractMatrix, AbstractVector, BroadcastStyle, IndexLinear, IndexStyle, OneTo, Slice,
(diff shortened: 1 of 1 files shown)


nsajko:
This is the mentioned warning:

```julia-repl
julia> using LazyArrays
Info Given LazyArrays was explicitly requested, output will be shown live 
WARNING: Constructor for type "MemoryLayout" was extended in `LazyArrays` without explicit qualification or import.
  NOTE: Assumed "MemoryLayout" refers to `ArrayLayouts.MemoryLayout`. This behavior is deprecated and may differ in future versions.`
  NOTE: This behavior may have differed in Julia versions prior to 1.12.
  Hint: If you intended to create a new generic function of the same name, use `function MemoryLayout end`.
  Hint: To silence the warning, qualify `MemoryLayout` as `ArrayLayouts.MemoryLayout` or explicitly `import ArrayLayouts: MemoryLayout`
Precompiling LazyArrays finished.
  6 dependencies successfully precompiled in 20 seconds. 3 already precompiled.
  1 dependency had output during precompilation:
┌ LazyArrays
│  [Output was shown above]
└
```

dlfivefifty:
I'm completely confused... I thought we already needed to explicitly import `MemoryLayout` to overload??

But I see we aren't doing this...

src/LazyArrays.jl
```diff
@@ -5,6 +5,7 @@ module LazyArrays
 
 using Base.Broadcast, LinearAlgebra, FillArrays, ArrayLayouts, SparseArrays
 import LinearAlgebra.BLAS
+import ArrayLayouts.MemoryLayout
```
dlfivefifty, line 8: I think adding it to the `import ArrayLayouts: ...` would make more sense

nsajko:
Release?

dlfivefifty:
If you want something registered you should bump the version number

PR #367, opened 2025-02-27
patch release v2.6.1

nsajko:


dlfivefifty:
It should be a patch release v2.6.1

PR #383, opened 2025-10-22
CachedArrayStyle for broadcasting with cached arrays

dlfivefifty:


DanielVandH:
Yeah I was just taking a look now, but I'll fix the tests first. That `AbstractLazyArrayStyle` approach seems good.

dlfivefifty:
Yes that should work! I can also add you as a contributor to this repos if you wanted to push directly to this PR, but I don't think it matters too much

DanielVandH:
One issue with the layout approach.. is this intentional?

```julia
julia> a = Accumulate(*, 1:5)
5-element Accumulate{Int64, 1, typeof(*), Vector{Int64}, UnitRange{Int64}}:
   1
   2
   6
  24
 120

julia> a isa LazyArrays.AbstractCachedArray
true

julia> MemoryLayout(a)
LazyLayout()

julia> a isa LazyArrays.AbstractCachedArray
true

julia> MemoryLayout(a) isa LazyArrays.CachedLayout
false

julia> MemoryLayout(a) isa LazyArrays.AbstractCachedLayout
ERROR: UndefVarError: `AbstractCachedLayout` not defined in `LazyArrays`
Suggestion: check for spelling errors or missing imports.
Stacktrace:
 [1] getproperty(x::Module, f::Symbol)
   @ Base .\Base_compiler.jl:47
 [2] top-level scope
   @ REPL[109]:1
```

Surely this should be a `CachedLayout`. Currently the only relevant method is for `CachedArray`s (not `Abstract`) where

```julia
cachedlayout(::Data, ::Array) where {Data,Array} = CachedLayout{Data,Array}()
MemoryLayout(C::Type{CachedArray{T,N,DAT,ARR}}) where {T,N,DAT,ARR} = cachedlayout(MemoryLayout(DAT), MemoryLayout(ARR))
```

For `AbstractCachedArray`s, what is the equivalent method? The type of `DAT` and `ARR` (the `data` and `array` fields) are not stored in the `AbstractCachedArray` type alone (only the element type and the dimensions). Should `AbstractCachedArray` just document an interface requiring a definition for `MemoryLayout`, and I have to just go add some for `Accumulate`, etc? Versioning would be a bit annoying here 

The approach I had in the PR previously would still work if we want to just avoid this issue for now, but maybe it's worth looking at anyway

DanielVandH:
Hm.. the layout change is a bit annoying. One of the downstream tests (just as an example, there's a lot of breakage downstream)

```julia
A = BandedMatrices._BandedMatrix(Vcat(2*Ones(1,∞), ((1 ./(1:∞)).+1/4)', Ones(1,∞)./3), ℵ₀, 1, 1)
Q, L = ql(A)          
@test Q' * zeros(∞,2) isa CachedArray
```

breaks now because of changing to a `GenericCachedLayout` instead of a `LazyLayout` for whatever `AbstractCachedArray` is involved in the `ql`, since now one of the methods that was previously used, namely

https://github.com/JuliaLinearAlgebra/InfiniteLinearAlgebra.jl/blob/f20670b97aaab548e080e5569afd8da54dfe1ed7/src/infql.jl#L500-L528

is not used and instead this finite version

https://github.com/JuliaLinearAlgebra/MatrixFactorizations.jl/blob/0630a3a37c18382020d143244d25fdacbcb0a61b/src/ql.jl#L350-L377

is used. Do you see any good way around this, beyond just going back to the type-based approach for now? Maybe it should be `AbstractCachedLayout <: AbstractLazyLayout` and these methods could instead be for `AbstractLazyLayouts` rather than `LazyLayout`s, although that's a lot of downstream changes.

dlfivefifty:
LGTM!

(Does that mean "Looks Good To Me" or "Let's Get This Merged"?)

I guess it's ready to merge?

DanielVandH:
Were you waiting for me to do the merging+tagging @dlfivefifty? Wasn't sure what your preference is for me making updates affecting the main branch. I think this is good to be merged & tagged. 

I think https://github.com/JuliaApproximation/SemiclassicalOrthogonalPolynomials.jl/pull/147 will then also be ready to be merged after a new tag assuming the tests still pass (they do locally, but can't test CI until a tag).

dlfivefifty:
Sorry I just missed your previous message that it’s ready

PR #386, opened 2025-10-29
Support resizedata! and cacheddata for Vcat

dlfivefifty:


DanielVandH:
I don't understand the original BoundsError for this test 

```julia
A = BandedMatrices._BandedMatrix(Vcat(BroadcastArray(exp, 1:5)', Ones(1,5)), 5, 1, 0)
x = Vcat([1,2], Zeros(3))
muladd!(1.0, A, x, 2.0, Vcat(zeros(2), Zeros(3)))
```

arising from this method

```julia
function _vcat_resizedata!(::Union{AbstractPaddedLayout, DualLayout{<:PaddedRows}}, B, m...)
    any(iszero,m) || Base.checkbounds(paddeddata(B), m...)
    B
end
```

which got changed with this `resizedata!` method. It must be important though since, if I bypass the BoundsError,

```julia
vcat(2-element Vector{Float64}, 3-element Zeros{Float64}):
  2.718281828459045
 15.7781121978613
   ⋅
   ⋅
   ⋅
```

but I think the correct answer should be 

```julia
julia> A * x
vcat(3-element Vector{Float64}, 2-element Zeros{Float64}):
  2.718281828459045
 15.7781121978613
  2.0
   ⋅
   ⋅
```

unless I'm mistaken?

PR #388, opened 2025-11-20
Fix dimensions for row view of a vcat

dlfivefifty:


DanielVandH:
Is this good to go now? Or did you want more work done on it?

PR #391, opened 2025-11-26
Various fixes for propagating CachedArrayStyle correctly

DanielVandH:
Now going to use this PR to get various other fixes for propagating CachedArrayStyle. Makes it easier for me when debugging but if it's preferred I can later break it up into smaller PRs. Once all the changes are in, I'll edit in all the new features here. Original PR text below.

---

This changes `Vcat` and `Hcat` to use the combined `BroadcastStyle` between a `LazyArrayStyle` and the combined style of its arguments. This fixes e.g. `Vcat` of cached arguments not being recognised as a `CachedArrayStyle`. 

The implementation needs to use some recursion to get the `BroadcastStyle` of the arguments via `tuple_type_broadcastlayout`, but basically it's just doing the same as what you'd get from `result_style(LazyArrayStyle{N}(), combine_style(A.args...))` where `A` is a `Vcat`/`Hcat`

---


DanielVandH:
Similar cases missing a `CachedArray` style also show up in e.g. BlockArrays and LazyBandedMatrices:

```julia-repl
julia> v = Accumulate(*, 1:10);

julia> (Base.BroadcastStyle ∘ typeof).((v, unitblocks(v), BlockBroadcastArray(vcat, unitblocks(v), unitblocks(v))))
(LazyArrays.CachedArrayStyle{1}(), BlockArrays.BlockedStyle{1}(), LazyArrays.LazyArrayStyle{1}())
```

So I might have to eventually go and fix these, assuming it would be appropriate that these all return `CachedArrayStyle{1}()`? Similarly, 

```julia-repl
julia> Base.BroadcastStyle(typeof(vec(Vcat(v', v')))) # interlaced
Base.Broadcast.DefaultArrayStyle{1}()
```

will get a fix (this last one is now fixed in https://github.com/JuliaLinearAlgebra/ArrayLayouts.jl/pull/268)

dlfivefifty:
I guess we don't need a `CachedBandedArrayStyle` because `CachedStyle` lowers to the data after resizing?

DanielVandH:
I think that's right. I wasn't too sure exactly about cases like that (and e.g. the block matrices like I mentioned in my previous comment), but I think I reasoned similarly that `CachedArrayStyle` is fine

DanielVandH:
This `QRCompactWYQ` type always seems to run into this issue. I'm not sure how to get around it properly this time, since there would need to be somewhere to call `broadcastable` earlier (which converts it into a `Matrix`), since by the time the calls get into using type information only it's too late. If `LinearAlgebra` just had 

```julia
BroadcastStyle(::Type{<:LinearAlgebra.QRCompactWYQ}) = DefaultArrayStyle{2}()
BroadcastStyle(::Type{<:LinearAlgebra.AdjointQ}) = DefaultArrayStyle{2}()
```

it would be fine, but can't do that on our side without piracy. I don't really understand why they don't already do this. 

Solutions like defining an internal `_BroadcastStyle` or overloading `combine_styles` seem like code smell. Have you had to deal with something similar before @dlfivefifty? An MWE for this is

```julia-repl
julia> using LazyArrays, LinearAlgebra

julia> B = LazyArrays.CachedArray(randn(3, 3)); Q = qr(randn(3, 3)).Q; collect(B); Q * B ≈ Q * B.data
ERROR: MethodError: no method matching ndims(::Type{LinearAlgebra.QRCompactWYQ{Float64, Matrix{Float64}, Matrix{Float64}}})
The function `ndims` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  ndims(::Type{Union{}}, Any...)
   @ Base abstractarray.jl:276
  ndims(::Type{<:Ref})
   @ Base refpointer.jl:104
  ndims(::Type{<:Applied{<:Any, typeof(*), Args}}) where Args
   @ LazyArrays c:\Users\djv23\.julia\dev\LazyArrays.jl\src\linalg\mul.jl:41
  ...

Stacktrace:
  [1] Base.Broadcast.BroadcastStyle(::Type{LinearAlgebra.QRCompactWYQ{Float64, Matrix{Float64}, Matrix{Float64}}})
    @ Base.Broadcast .\broadcast.jl:103
  [2] tuple_type_broadcastlayout
    @ c:\Users\djv23\.julia\dev\LazyArrays.jl\src\lazyconcat.jl:445 [inlined]
  [3] applybroadcaststyle(::Type{ApplyArray{Float64, 2, typeof(*), Tuple{…}}}, ::ApplyLayout{typeof(*)})
    @ LazyArrays c:\Users\djv23\.julia\dev\LazyArrays.jl\src\lazyapplying.jl:334
  [4] Base.Broadcast.BroadcastStyle(M::Type{ApplyArray{Float64, 2, typeof(*), Tuple{…}}})
    @ LazyArrays c:\Users\djv23\.julia\dev\LazyArrays.jl\src\lazyapplying.jl:336
  [5] combine_styles(c::ApplyArray{Float64, 2, typeof(*), Tuple{LinearAlgebra.QRCompactWYQ{…}, CachedArray{…}}})
    @ Base.Broadcast .\broadcast.jl:433
  [6] combine_styles(c1::ApplyArray{Float64, 2, typeof(*), Tuple{…}}, c2::Matrix{Float64})
    @ Base.Broadcast .\broadcast.jl:438
  [7] broadcasted
    @ .\broadcast.jl:1353 [inlined]
  [8] broadcast_preserving_zero_d
    @ .\broadcast.jl:882 [inlined]
  [9] -(A::ApplyArray{Float64, 2, typeof(*), Tuple{LinearAlgebra.QRCompactWYQ{…}, CachedArray{…}}}, B::Matrix{Float64})
    @ Base .\arraymath.jl:8
 [10] isapprox(x::ApplyArray{…}, y::Matrix{…}; atol::Int64, rtol::Float64, nans::Bool, norm::typeof(norm))
    @ LinearAlgebra C:\Users\djv23\.julia\juliaup\julia-1.12.1+0.x64.w64.mingw32\share\julia\stdlib\v1.12\LinearAlgebra\src\generic.jl:1982
 [11] isapprox(x::ApplyArray{Float64, 2, typeof(*), Tuple{LinearAlgebra.QRCompactWYQ{…}, CachedArray{…}}}, y::Matrix{Float64})
    @ LinearAlgebra C:\Users\djv23\.julia\juliaup\julia-1.12.1+0.x64.w64.mingw32\share\julia\stdlib\v1.12\LinearAlgebra\src\generic.jl:1978
 [12] top-level scope
    @ REPL[100]:1
Some type information was truncated. Use `show(err)` to see complete types.
```

dlfivefifty:
I think

> defining  an internal _BroadcastStyle

is probably the best way around this?

I have played around in the past with making a type so that Q's are `<: AbstractMatrix` (I think `AbstractQ` is kinda poorly thought out btw):

https://github.com/JuliaLinearAlgebra/MatrixFactorizations.jl/pull/69

DanielVandH:
I'm running into a bit of an annoying problem. The problem stems from trying to support `cacheddata(::ApplyArray)` and `cacheddata(::BroadcastArray)`, simply rewrapping the arguments in `maybe_cacheddata` calls back into another `(Apply/Broadcast)Array`. This is one of the last things we might need for our timing stuff since we're losing a lot of the `CachedArrayStyle` to `ApplyArray`/`BroadcastArray`s currently.

#

First I'll just explain a new function `conforming_resize!`, then some problems with it.

To properly get the `CachedArrayStyle` to propagate correctly, I needed to add support for `cacheddata` on `ApplyArray` and `BroadcastArray`, which creates an `ApplyArray`/`BroadcastArray` over the same arguments except with their cached data. This will also have to make sure all the `cacheddata` calls on the arguments of those arrays match all the other arguments (since e.g., `BroadcastVector(+, [1, 2], [1, 2, 3])` as an output from a `cacheddata` when the arguments have differerent datasizes wouldn't make sense). The logic for this is implemented (which, admittedly, looks quite rough for the moment) in `conforming_resize!` here

https://github.com/JuliaArrays/LazyArrays.jl/blob/78b3c33486c59ec909a9a64051b59c46ac8674d5/src/cache.jl#L622-L701

and called e.g. for `ApplyArray` here 

https://github.com/JuliaArrays/LazyArrays.jl/blob/78b3c33486c59ec909a9a64051b59c46ac8674d5/src/lazyapplying.jl#L231-L235

The code for `conforming_resize!` is a bit complex because of issues I was having with e.g. recognising that `cache(rand(1, 10))'` is a vector but its size looks more like a matrix. You can look at the tests here 

https://github.com/JuliaArrays/LazyArrays.jl/blob/78b3c33486c59ec909a9a64051b59c46ac8674d5/test/cachetests.jl#L630-L713

to try and see all the cases I was supporting so far if you did want to understand why it has to be so complex (not that there isn't probably a better way, which hopefully we can discuss).

#

Some problems:

1. This approach doesn't really work in the case of e.g. 
```julia-repl
julia> A, b = rand(2, 2), rand(2);

julia> M = BroadcastMatrix(*, cache(A), cache(b));

julia> LazyArrays.cacheddata(M)
ERROR: ArgumentError: Cannot conform arrays with incompatible dimensions: ((2, 2), (2,))
Stacktrace:
 [1] conforming_resize!
   @ c:\Users\djv23\.julia\dev\LazyArrays.jl\src\cache.jl:698 [inlined]
 [2] cacheddata(A::BroadcastMatrix{Float64, typeof(*), Tuple{Matrix{Float64}, Vector{Float64}}})
   @ LazyArrays c:\Users\djv23\.julia\dev\LazyArrays.jl\src\lazybroadcasting.jl:127
 [3] top-level scope
   @ REPL[167]:1
```
and I'm not seeing the way to fix that currently.

2. The current implementation tries to resize to an infinite arrays in the case that one of the arguments is infinite and not a cached array:
```julia-repl
julia> A = Vcat((1:∞)', cache(1:∞)');

julia> B = view(A, :, 1:10);

julia> LazyArrays.cacheddata(B) 
ERROR: ArgumentError: Cannot create infinite Array
Stacktrace:
  [1] Vector{Int64}(::UndefInitializer, ::Tuple{Infinities.InfiniteCardinal{0}})
    @ InfiniteArrays C:\Users\djv23\.julia\packages\InfiniteArrays\v6xnl\src\infarrays.jl:19
  [2] Vector{Int64}(x::LazyArrays.CachedArray{Int64, 1, Vector{Int64}, Zeros{Int64, 1, Tuple{InfiniteArrays.OneToInf{…}}}})
    @ Base .\array.jl:621
  [3] convert(::Type{Vector{Int64}}, a::LazyArrays.CachedArray{Int64, 1, Vector{Int64}, Zeros{Int64, 1, Tuple{…}}})
    @ Base .\array.jl:614
  [4] setproperty!(x::LazyArrays.CachedArray{…}, f::Symbol, v::LazyArrays.CachedArray{…})
    @ Base .\Base_compiler.jl:57
  [5] _vec_resizedata!(B::LazyArrays.CachedArray{…}, n::Infinities.InfiniteCardinal{…})
    @ LazyArrays c:\Users\djv23\.julia\dev\LazyArrays.jl\src\cache.jl:254
  [6] resizedata!(::DenseColumnMajor, ::LazyArrays.LazyLayout, B::LazyArrays.CachedArray{…}, n::Infinities.InfiniteCardinal{…})
    @ LazyArrays c:\Users\djv23\.julia\dev\LazyArrays.jl\src\cache.jl:266
  [7] resizedata!(B::LazyArrays.CachedArray{…}, mn::Infinities.InfiniteCardinal{…})
    @ LazyArrays c:\Users\djv23\.julia\dev\LazyArrays.jl\src\cache.jl:229
  [8] resizedata!(A::Adjoint{Int64, LazyArrays.CachedArray{…}}, m::Int64, n::Infinities.InfiniteCardinal{0})
    @ LazyArrays c:\Users\djv23\.julia\dev\LazyArrays.jl\src\cache.jl:237
  [9] _resize_each!
    @ c:\Users\djv23\.julia\dev\LazyArrays.jl\src\cache.jl:711 [inlined]
 [10] _resize_each!
    @ c:\Users\djv23\.julia\dev\LazyArrays.jl\src\cache.jl:712 [inlined]
 [11] conforming_resize!(args::Tuple{Adjoint{…}, Adjoint{…}})
    @ LazyArrays c:\Users\djv23\.julia\dev\LazyArrays.jl\src\cache.jl:702
 [12] cacheddata(A::ApplyArray{Int64, 2, typeof(vcat), Tuple{Adjoint{…}, Adjoint{…}}})
    @ LazyArrays c:\Users\djv23\.julia\dev\LazyArrays.jl\src\lazyapplying.jl:233
 [13] cacheddata(V::SubArray{Int64, 2, ApplyArray{…}, Tuple{…}, false})
    @ LazyArrays c:\Users\djv23\.julia\dev\LazyArrays.jl\src\cache.jl:617
 [14] top-level scope
    @ REPL[172]:1
Some type information was truncated. Use `show(err)` to see complete types.
```
this example fails because it wants to create `cacheddata(A)` which in turn wants to expand the `cacheddata` for the second argument to infinity since the first argument is a non-cached argument with infinite length. This is the most difficult part for me. 

There are a few ideas I have, like trying to reimplement `cacheddata` specifically for views of `BroadcastArray`/`ApplyArray`. This is all starting to feel very over-engineered however, so I wanted to see if you had any ideas @dlfivefifty before I add even more code that probably wouldn't hold up to even more intense debugging anyway.

PR #411, opened 2026-06-26
Fix BlockHcat sub-indexing

dlfivefifty:


dlfivefifty:
@copilot why didn't codecov run?

dlfivefifty:
@copilot it says No report found to compare against

PR #418, opened 2026-08-03
Support sortedunion for eg Vcat(1, 3:10)

dlfivefifty:
This works around an issue with infinite block vectors downstream


dlfivefifty:
@copilot get diff coverage to 100%

PR #420, opened 2026-09-03
Use back substition for uppertriangular banded \ padded

dlfivefifty:


dlfivefifty:
@copilot increase code coverage

